PR #31, opened 2026-03-25
feat(napari-plugin): load video side-by-side to the fUSI scan

FelipeCybis:
## Summary

Adds a Video panel to the napari plugin for loading one or more behavioral videos alongside an fUSI scan. Each video becomes its own image layer, overlaid on the reference scan, and is shown in its own grid cell so videos and the scan sit side by side.

#### Blockers for merging
- [x] User guide in the docs
- [x] Usable public dataset for examples and user guide

### What you can do

  - **Load multiple videos**. The Video accordion panel has a "Loaded videos" list; each Add video click appends a new layer. Videos can be removed individually via the list + Remove selected button.
  - **Frame-synchronized playback**. Videos share the fUSI time axis — scrubbing the time slider moves every loaded video to the matching frame. The time scale is frame_step / fps, so napari's world coordinate stays in physical
  seconds.
  - **Frame step subsampling**. A spin box shows every N-th frame across all loaded videos, useful for long/heavy recordings.).
  - **CLI support**. confusius <scan> --video <path> loads a video alongside the scan at startup.

### Quick example

https://github.com/user-attachments/assets/0914f896-dee8-4b27-9d17-30367d21f37a

### Take a tour

image

image

image

image


FelipeCybis:
@sdiebolt I think I maxxed out what I could do in terms of perf. I will work on docs and explaining the widget now, but can you start testing it?

The only philosophical question remaining is: what to do with multiple videos?
Currently, only 1 is allowed. If the user loads another one, the current layer is removed and another one is added.
As we discussed, the must would be having multiple canvas https://github.com/napari/napari/issues/5348.
I am currently enabling grid mode when I add the video layer. New layers overlay the scan layer (if we want to add labels and such). Should we hijack the grid mode logic to add a new video layer under an existing video layer? (maybe that is not a bad idea, actually).

sdiebolt:
Nice one, I'll start testing! About the multi-video stuff, I have some recordings with multiple videos associated with the same fUSI recording, so I'm also interested in that. Can't we just allow loading multiple video, keep track of how many videos are loaded, and adjust the grid width accordingly?

FelipeCybis:
> Nice one, I'll start testing! About the multi-video stuff, I have some recordings with multiple videos associated with the same fUSI recording, so I'm also interested in that. Can't we just allow loading multiple video, keep track of how many videos are loaded, and adjust the grid width accordingly?


https://github.com/user-attachments/assets/0914f896-dee8-4b27-9d17-30367d21f37a

Almost there!

FelipeCybis:
@sdiebolt this is good to be reviewed!

sdiebolt:
> @sdiebolt this is good to be reviewed!

Nice one! I'll review ASAP (most probably next monday/tuesday, sooner if I'm able), I'm currently at the Gozzi lab and have full days of meetings/workshops, so bandwidth is a bit tight :sweat_smile:

sdiebolt:
@FelipeCybis I found some time to review this, and it is awesome! This module is really awesome, it integrates super well with the rest of the napari modules and runs sooooo smoothly! Huge step toward the beta-release :rocket:

src/confusius/_napari/_qc/_plots.py
```diff
@@ -375,6 +376,15 @@ def set_time_cursor(self, time_val: float) -> None:
 
     def _flush_cursor(self) -> None:
         """Perform the actual blit for the current cursor position."""
+        if self._flushing_cursor:
+            return
+        self._flushing_cursor = True
+        try:
```
sdiebolt, line 386: Maybe this would warrant a comment to explain why it's needed? Are there cases where showing the cursor fails?

sdiebolt (approved):
LGTM! Huge work!

src/confusius/_napari/_qc/_plots.py
FelipeCybis: Indeed, this is not obvious. The problem this is solving is that now we are redrawing the cursor asynchronously, and now nothing is preventing a new event to ask for flushing the cursor while the previous one hasn't finished. This can happen when the animation is running and the user clicks on the plot to go to the time position, for example. I put a comment.

PR #37, opened 2026-04-02
feat(glm): add a GLM module

### LINKED ISSUE #15: Scaffold `glm` module

Create the `confusius.glm` module for general linear model analysis of fUSI data,
following the nilearn design pattern.

  ## Planned scope (alpha)

  - `FirstLevelModel`: fit a voxel-wise GLM given a design matrix (stimuli
    convolved with an HRF, confound regressors) and return contrast maps as
    `xr.DataArray`.
  - `SecondLevelModel`: fit a group-level GLM over a collection of first-level
    contrast maps (e.g., fixed-effects or mixed-effects across subjects/sessions).
  - Design matrix helpers: build first-level design matrices from stimulus
    paradigms (onset/duration/amplitude), HRF convolution, and confound columns.

sdiebolt:
This PR implements a voxel-wise General Linear Model for fUSI time-series analysis, following a two-pass OLS → AR whitening approach.

### First-level analysis
- `FirstLevelModel` sklearn-style estimator with multi-run fixed-effects combination, supporting OLS and AR(p) noise models.
- Per-voxel AR whitening: each voxel uses its own AR coefficients.
- Design matrix builder (`make_first_level_design_matrix`) with HRF convolution, drift regressors, and confound support.
- HRF models: Glover, SPM, Claron *et al.* (2021), Verhoef *et al.* (2025), FIR, plus arbitrary callables matching the `HRFModel` protocol.
- `Contrast` result container with t/F statistics, p-values, and z-scores; arithmetic operators for fixed-effects combination across runs.

### Second-level analysis
- `SecondLevelModel` sklearn-style estimator for group-level analyses, accepting either a list of fitted `FirstLevelModel` objects or a list of spatial contrast maps.
- `make_second_level_design_matrix` helper for one-sample and confound-augmented group designs.

### Output and integration
- `compute_contrast` returns spatial `xarray.DataArray` maps with attribute propagation (input attrs equal across runs/maps are preserved on the output).
- `output_type` selects between `"zscore"`, `"statistic"`, `"pvalue"`, `"effect"`, `"variance"` (scipy/statsmodels-style attribute names on the underlying `Contrast`).

Closes #15.


src/confusius/glm/_design.py
```diff
@@ -0,0 +1,869 @@
...
+    """
+    return interp1d(volume_times_high_res, high_res_regressor)(volume_times).T
+
+
+def _hrf_kernel(
+    hrf_model: str | None,
+    dt: float,
+    oversampling: int = 50,
+    fir_delays: list[int] | None = None,
+) -> list[npt.NDArray[np.floating]]:
```
FelipeCybis: It would be cool to have this accepting a callable as well, like nilearn, I mean, `hrf_model` that could be a callable in `FirstLevelModel`. Maybe we can create a `Protocol` type for it.

FelipeCybis: This would ease the implementation of new hrf models such as Claron et al. 2021 (that we could implement out of the bat here (I can do that) or in another PR).

src/confusius/glm/_design.py
```diff
@@ -0,0 +1,869 @@
...
+    dt = float(intervals[0])
+
+    if dt <= 0:
+        raise ValueError("volume_times must be strictly increasing.")
+    if not np.allclose(intervals, dt, rtol=1e-5, atol=1e-8):
+        raise ValueError(
+            "Volume times must be uniformly spaced to compute sampling interval. "
+            f"Found varying intervals: min={intervals.min():.6f}, "
+            f"max={intervals.max():.6f}."
+        )
```
FelipeCybis: Maybe this is too strict? This makes me think that having an `interpolate_timings` function (more general then `correct_slice_timings` could be a good idea. I could the mention that using that could be a good idea. 

Either way, I am not sure if we would really want a raise here, maybe `1e-5 < rtol < 1e-3` we could warn and use `dt` or `(volume_times[-1] - volume_times[0])/(n_frames-1)` and raise if `rtol < 1e-3`?

src/confusius/glm/_design.py
FelipeCybis: Both implemented in https://github.com/confusius-tools/confusius/pull/37/commits/a9837504963b7036d5c407301c0cd0604277a61c

src/confusius/glm/_contrasts.py
```diff
@@ -0,0 +1,373 @@
...
+        """Initialize contrast container."""
+        if variance.ndim != 1:
+            raise ValueError(f"variance must be 1D, got shape {variance.shape}")
+
+        if effect.ndim not in (1, 2):
+            raise ValueError(f"effect must be 1D or 2D, got shape {effect.shape}")
+
+        self.effect: npt.NDArray[np.floating] = effect
+        self.variance = variance
+        self.dof = float(dof)
```
sdiebolt:
```suggestion
        self.dof = dof
```

src/confusius/glm/_contrasts.py
```diff
@@ -0,0 +1,373 @@
...
+        if dim is None:
+            self.dim = effect.shape[0] if effect.ndim == 2 else 1
+        else:
+            self.dim = dim
+
+        if self.dim > 1 and stat_type == "t":
+            warnings.warn(
+                f"stat_type='t' is incompatible with a {self.dim}-dimensional contrast; "
+                "switching to 'F'.",
+                UserWarning,
```
sdiebolt:
```suggestion
```

src/confusius/glm/_contrasts.py
```diff
@@ -0,0 +1,373 @@
...
+        if self.p_value_ is None or self.baseline != baseline:
+            self.p_value(baseline)
+
+        if self.one_minus_pvalue_ is None or self.baseline != baseline:
+            self.one_minus_pvalue(baseline)
+
+        assert self.p_value_ is not None
+        assert self.one_minus_pvalue_ is not None
+        self.z_score_ = _pvalues_to_zscore(self.p_value_, self.one_minus_pvalue_)
+        return self.z_score_
```
sdiebolt: These methods were inspired by Nilearn, but maybe we should make them properties?

sdiebolt:
@FelipeCybis I'm almost done with the revisions (only some tests to verify and a few docstrings to touchup) and we should be done to merge! I expect to be able to merge this by the end of today :rocket:

src/confusius/qc/dvars.py
```diff
@@ -222,15 +180,13 @@ def compute_dvars(
...
-        [_ar1_yule_walker(signals_np[:, i]) for i in range(signals_np.shape[1])]
-    )
+    ar1, _ = estimate_ar_coeffs(signals_np, order=1)
 
     # For an AR(1) process, the variance of temporal differences is:
     # Var(diff) = Var(x) * (2 * (1 - rho)) = 2 * (1 - rho) * sigma^2
     # where rho is the AR(1) coefficient and sigma^2 is the process variance. See
     # Nichols (2017).
-    diff_sdhat = np.sqrt(2 * (1 - ar1)) * signals_sd
+    diff_sdhat = np.sqrt(2 * (1 - ar1.squeeze())) * signals_sd
```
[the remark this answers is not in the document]
sdiebolt: Closing without action: with the current Yule-Walker estimator, the AR(1) coefficient is `r1/r0 = (Σ x_i x_{i+1}) / (Σ x_i²)`, which is bounded `|rho| ≤ 1` by Cauchy-Schwarz. Even at `rho = 1`, `sqrt(2·(1−rho)) = 0` — finite, never NaN. The clip in the previous local AR(1) estimator was prophylactic; it is not needed with this estimator.

PR #47, opened 2026-04-10
fix(napari): track viewer resizing in guided tour overlay

FelipeCybis:
The tour scrim, spotlight cutout, and tooltip were positioned once per step and never updated, so resizing the napari viewer left the overlay stranded at its original geometry. `GuidedTour` now installs an event filter on the top-level window and recomputes geometry on `QEvent.Resize`.

Also prevent the "Take a Tour" button from spawning concurrent tours: `ConfUSIusWidget` tracks the active tour and clears the reference via the `GuidedTour.finished` signal so a fresh tour can start after skip/finish.

Closes #45 

https://github.com/user-attachments/assets/8345f3fb-b845-4d46-be7b-89357e185a16

### DIFF
--- a/src/confusius/_napari/_tour.py
+++ b/src/confusius/_napari/_tour.py
@@ -469,7 +469,7 @@ class GuidedTour(QObject):
         self._apply_geometry(step, target)
 
     def _apply_geometry(self, step: TourStep, target: QWidget) -> None:
-        """Resize the scrim and place the spotlight + tooltip for ``step``."""
+        """Resize the scrim and place the spotlight + tooltip for `step`."""
         # Keep overlay geometry in sync with the window (handles resizes).
         self._overlay.setGeometry(self._window.rect())
 
--- a/src/confusius/_napari/_tour.py
+++ b/src/confusius/_napari/_tour.py
@@ -41,8 +41,8 @@ class TourStep:
     body : str
         Longer explanatory text.
     anchor : str
-        Preferred tooltip placement relative to the target: ``"right"``,
-        ``"left"``, ``"above"``, or ``"below"``.
+        Preferred tooltip placement relative to the target: `"right"`,
+        `"left"`, `"above"`, or `"below"`.
     spotlight_rect : Callable[[], QRect | None] | None
         Optional callable returning a custom spotlight rectangle in window
         coordinates. Use this when a step should highlight multiple related
@@ -351,6 +351,8 @@ class GuidedTour(QObject):
 
     def start(self) -> None:
         """Show the overlay and display the first step."""
+        if self._active:
+            return
         self._overlay.setGeometry(self._window.rect())
         self._overlay.show()
         self._overlay.raise_()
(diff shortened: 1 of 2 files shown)


src/confusius/_napari/_tour.py
```diff
@@ -370,6 +382,35 @@ def close_tour(self) -> None:
...
 
+    # -- Event handling ------------------------------------------------------
+
+    def eventFilter(self, watched: QObject | None, event: QEvent | None) -> bool:  # type: ignore[invalid-method-override]  # noqa: N802
+        """Reposition the overlay when the watched window is resized.
+
+        Parameters
+        ----------
+        watched : qtpy.QtCore.QObject or None
+            The object the event was sent to; only ``self._window`` is acted
```
sdiebolt:
```suggestion
            The object the event was sent to; only `self._window` is acted
```

src/confusius/_napari/_tour.py
```diff
@@ -410,6 +451,25 @@ def _position_step(self, index: int, generation: int) -> None:
...
+        if not (0 <= self._current < len(self._steps)):
+            return
+        step = self._steps[self._current]
+        target = step.target()
+        if target is None or not target.isVisible():
+            return
+        self._apply_geometry(step, target)
+
+    def _apply_geometry(self, step: TourStep, target: QWidget) -> None:
+        """Resize the scrim and place the spotlight + tooltip for ``step``."""
```
sdiebolt:
```suggestion
        """Resize the scrim and place the spotlight + tooltip for `step`."""
```

sdiebolt (commented):
Super minor doc style fixes: since we use markdown for docstrings, there shouldn't be double backticks. We should keep that consistent otherwise LLMs will start using double backticks everywhere (they're already prone to do it given that Sphinx is used in the vast majority of packages).

sdiebolt (approved):
LGTM, thanks for handling this! :rocket:

FelipeCybis:
> Super minor doc style fixes: since we use markdown for docstrings, there shouldn't be double backticks. We should keep that consistent otherwise LLMs will start using double backticks everywhere (they're already prone to do it given that Sphinx is used in the vast majority of packages).

Oh, I didn't know that! Thanks!

src/confusius/_napari/_tour.py
```diff
@@ -354,13 +356,23 @@ def start(self) -> None:
         self._overlay.raise_()
         self._tooltip.show()
         self._tooltip.raise_()
+        self._active = True
+        # Watch the top-level window for resize events so the scrim, spotlight,
+        # and tooltip stay aligned when the napari window or its docks change
+        # size while the tour is active.
+        self._window.installEventFilter(self)
+        self._event_filter_installed = True
         # Defer one event-loop iteration so the tooltip completes its initial
```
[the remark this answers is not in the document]
FelipeCybis: @copilot apply changes based on [this feedback](https://github.com/confusius-tools/confusius/pull/47#discussion_r3065126946)

sdiebolt:
> Oh, I didn't know that! Thanks!

I'ts one minor disadvantage of Zensical vs. Sphinx + numpydoc: the latter allowed automatic cross-links for functions/methods/classes/etc. when using single backticks. But in markdown, double backticks are meaningless. Although the plugin system for Zensical is almost ready, so I expect we'll soon have easier cross-linking features!

PR #61, opened 2026-04-28
fix(extract): support integer masks in unmask

### LINKED ISSUE #59: `extract.unmask` fails with integer masks accepted by `extract_with_mask`
### ConfUSIus version

v0.0.1-a25

### Python version

3.13

### Operating system

Linux

### Installation

using uv

### Description

`confusius.extract.extract_with_mask` accepts integer single-label masks (0 = background, one non-zero region id = foreground). Passing the same mask to `confusius.extract.unmask` can fail with a shape mismatch error. This suggests `unmask` is using flattened integer mask values directly for NumPy indexing, so NumPy treats them as integer indices instead of a boolean mask.

### Steps to reproduce

```python
import numpy as np
import xarray as xr
import confusius as cf

# Synthetic data: (time, z, y, x)
data = xr.DataArray(
    np.random.default_rng(0).normal(size=(2, 2, 2, 3)),
    dims=["time", "z", "y", "x"],
)

# Integer single-label mask: 0 = background, 1 = foreground
mask_int = xr.DataArray(
    np.array(
        [
            [[1, 0, 1], [0, 0, 1]],
            [[0, 1, 0], [1, 0, 0]],
        ],
        dtype=np.int32,
    ),
    dims=["z", "y", "x"],
)

signals = cf.extract.extract_with_mask(data, mask_int)
print("signals shape:", signals.shape)  # (2, 5)

# Fails
cf.extract.unmask(signals, mask=mask_int)
```

### Additional output

_No response_

sdiebolt:
## Summary
- treat non-zero mask values as foreground in `extract.unmask` by using boolean indexing (`mask != 0`)
- validate `space` length against `np.count_nonzero(mask)` for both NumPy and DataArray signal inputs
- add regression tests covering integer-mask roundtrip behavior for DataArray and NumPy signals

Close #59

### DIFF
--- a/src/confusius/extract/reconstruction.py
+++ b/src/confusius/extract/reconstruction.py
@@ -94,9 +94,10 @@ def unmask(
     >>> spatial_pose.dims
     ("component", "pose", "z", "y", "x")
     """
-    if isinstance(signals, np.ndarray):
-        n_voxels_mask = int(mask.sum().values)
+    mask_values = mask.values
+    n_voxels_mask = int(np.count_nonzero(mask_values))
 
+    if isinstance(signals, np.ndarray):
         if signals.shape[-1] != n_voxels_mask:
             raise ValueError(
                 f"Last dimension of signals ({signals.shape[-1]}) doesn't match "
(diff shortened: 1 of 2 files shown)


sdiebolt:
@FelipeCybis This should be good to review and merge!

FelipeCybis (approved):
LGTM!

PR #62, opened 2026-04-28
fix(extract): stabilize mask extraction roundtrips

### LINKED ISSUE #60: `extract_with_mask(...).unstack("space")` can rebuild a non-canonical volume grid
## Problem

`extract_with_mask` can return a `space` coordinate / MultiIndex that leads `signals.unstack("space")` to rebuild a volume on a non-canonical spatial grid.

In practice, the extracted values are correct, but the reconstructed `x/y/z` coordinate arrays coming out of `unstack("space")` can be ordered in a way that makes downstream plotting look scrambled.

This issue is about the `space` index produced by `extract_with_mask`, not only about plotting.

## Repro

A reliable real-world path is:

```python
import confusius as cf

pwd = cf.load("/home/sdiebolt/Downloads/4Dscan_2_good_fus3D.source.scan")
pwd = cf.multipose.consolidate_poses(pwd).compute()

viewer, layer = cf.plotting.draw_napari_labels(pwd.mean("time"), gamma=0.5)
brain_mask = cf.plotting.labels_from_layer(layer, pwd)

signals = cf.extract.extract_with_mask(pwd, brain_mask)
u = signals.unstack("space")
```

On affected code, `u` does not reconstruct on a canonical spatial grid, and plotting `u` can look scrambled unless coordinates are sorted before rendering.

A synthetic repro also exists when the selected `space` index is rebuilt from a sparse masked subset in a way that preserves a non-canonical internal MultiIndex state.

## Expected

`extract_with_mask(...).unstack("space")` should reconstruct a volume whose spatial coordinates are canonical / suitable for direct plotting, and should be equivalent to `unmask` on the recoverable voxels.

## Actual

The extracted voxel values are correct, but `unstack("space")` can rebuild a volume with a non-canonical `space` index / coordinate grid, so plots of the unstacked result look scrambled.

## Fix direction

- After masking, rebuild the selected `space` index from explicit spatial coordinates in canonical order.
- Keep plotting-side sorting as a separate robustness layer, since users may also plot non-canonical arrays from other sources.

sdiebolt:
## Summary
- rebuild extracted `space` indexing so `extract_with_mask(...).unstack("space")` produces a stable reduced-grid volume
- keep `unmask` compatible with both DataArray and ndarray signal inputs after extraction
- sort coordinates during plotting as a separate robustness layer for arrays coming from other sources

Close #60

### DIFF
--- a/src/confusius/extract/mask.py
+++ b/src/confusius/extract/mask.py
@@ -96,7 +96,7 @@ def extract_with_mask(data: xr.DataArray, mask: xr.DataArray) -> xr.DataArray:
         return data.isel(space=mask_flat)
 
     data_flat = data.stack(space=spatial_dims)
-    mask_flat = mask_aligned.stack(space=spatial_dims).astype(bool)
+    mask_flat = mask_aligned.values.ravel().astype(bool)
     # Rebuild the space index from the selected voxel coordinates so unstack() uses the
     # reduced grid implied by the extracted mask.
-    return data_flat.isel(space=mask_flat.values).set_xindex(spatial_dims)
+    return data_flat.isel(space=mask_flat).set_xindex(spatial_dims)
--- a/src/confusius/extract/mask.py
+++ b/src/confusius/extract/mask.py
@@ -96,5 +96,7 @@ def extract_with_mask(data: xr.DataArray, mask: xr.DataArray) -> xr.DataArray:
         return data.isel(space=mask_flat)
 
     data_flat = data.stack(space=spatial_dims)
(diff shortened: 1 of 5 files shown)


sdiebolt:
@FelipeCybis Please wait before reviewing this. I think I can modify `extract_with_mask` and `unmask` to actually output signals in an order that would make `unstack` output properly sorted coordinates. But I'm waiting until #60 is merged before doing that.

sdiebolt:
@FelipeCybis This is good to review!

FelipeCybis (approved):

PR #69, opened 2026-04-29
fix(extract): align near-equal mask coordinates

### LINKED ISSUE #63: `extract_with_mask` selects all voxels when mask coords are near-equal but not exact
### ConfUSIus version

v0.0.1-a25

### Python version

3.13

### Operating system

Linux

### Installation

using uv

### Description

`extract_with_mask` can silently extract every voxel when mask coordinates differ slightly from data coordinates (e.g. after NIfTI roundtrip). If coordinates are not exactly alignable, extraction should fail with a clear error (or use an explicit tolerant alignment mode), not silently return full-volume extraction.

### Steps to reproduce

```python
import numpy as np
import xarray as xr
import confusius as cf

data = xr.DataArray(
    np.random.default_rng(0).normal(size=(3, 2, 3, 4)),
    dims=["time", "z", "y", "x"],
    coords={
        "z": np.array([0.0, 1.0]),
        "y": np.array([0.0, 1.0, 2.0]),
        "x": np.array([0.0, 1.0, 2.0, 3.0]),
    },
)

z_mask = np.array([0.0, np.nextafter(1.0, 2.0)])
y_mask = np.array([0.0, np.nextafter(1.0, 2.0), np.nextafter(2.0, 3.0)])
x_mask = np.array([0.0, np.nextafter(1.0, 2.0), np.nextafter(2.0, 3.0), np.nextafter(3.0, 4.0)])

mask = xr.DataArray(
    np.array(
        [
            [[1, 0, 1, 0], [0, 1, 0, 0], [1, 0, 0, 0]],
            [[0, 1, 0, 1], [0, 0, 1, 0], [0, 0, 0, 1]],
        ],
        dtype=np.int32,
    ),
    dims=["z", "y", "x"],
    coords={"z": z_mask, "y": y_mask, "x": x_mask},
)

signals = cf.extract.extract_with_mask(data, mask)
print("expected non-zero:", int((mask.values != 0).sum()))  # 8
print("extracted space:", signals.sizes["space"])          # 24 (full volume)
```

### Additional output

_No response_

sdiebolt:
## Summary
- snap validated mask coordinates onto the data template before `reindex_like` so near-equal voxel grids align exactly
- prevent silent full-volume extraction when masks survive tolerant validation but fail exact label alignment
- add a regression test covering near-equal coordinates from a NIfTI-like roundtrip

Close #63.

### DIFF
--- a/src/confusius/extract/mask.py
+++ b/src/confusius/extract/mask.py
@@ -1,5 +1,6 @@
 """Extraction of signals using boolean masks."""
 
+import numpy as np
 import xarray as xr
 
 from confusius.validation import validate_mask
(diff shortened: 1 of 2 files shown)


sdiebolt:
@FelipeCybis Good to review!

FelipeCybis (approved):
LGTM!

PR #72, opened 2026-04-29
fix(io): handle scalar-time NIfTI metadata and bool mask export

### LINKED ISSUE #71: NIfTI save/load edge cases for scalar time coords and boolean masks
## Summary
Saving and reloading 3D snapshots with scalar `time` metadata currently shows two edge-case failures:

1. Saving a 3D array with a scalar `time` coordinate can crash timing inference (`IndexError: invalid index to scalar variable`).
2. Saving boolean mask payloads to NIfTI can fail with nibabel `HeaderDataError` because NIfTI does not support `bool` data dtype.

There is also a related roundtrip ergonomics issue where timing metadata for 3D snapshots can end up in top-level attrs instead of scalar coordinates.

## Reproduction
```python
import confusius as cf
pwd = cf.load("/home/sdiebolt/Downloads/4Dscan_2_good_fus3D.source.scan")
pwd = cf.multipose.consolidate_poses(pwd).compute()
mask = pwd.isel(time=0) > 100
cf.save(mask, "mask.nii")
```

## Expected
- Scalar `time` metadata should serialize/deserialize without crashing.
- Boolean masks should save to a NIfTI-compatible dtype.
- Scalar timing semantics should roundtrip consistently for 3D snapshots.

## Actual
- `IndexError` in timing inference when indexing scalar `time` values.
- `HeaderDataError` for boolean payload dtype.
- Scalar timing metadata may be awkwardly represented on reload.

sdiebolt:
## Summary
- fix scalar `time` save-path handling by normalizing timing arrays to 1D before TR inference, preventing crashes on 3D arrays with scalar time metadata
- write boolean NIfTI payloads as `uint8` so mask exports are nibabel-compatible
- improve scalar-time roundtrip ergonomics for 3D NIfTI payloads by reconstructing scalar `time` (and scalar `slice_time` when present) from sidecar metadata
- support exporting 1D spatial `slice_time` coordinates for scalar-time snapshots in addition to existing 2D `(time, spatial_dim)` export
- add targeted regression tests for scalar-time save/load, 1D slice-time export/load, and bool mask export

Fixes #71

### DIFF
--- a/src/confusius/bids/__init__.py
+++ b/src/confusius/bids/__init__.py
@@ -27,6 +27,8 @@ Examples
 """
 
 from confusius.bids.coordinates import (
+    DIM_TO_SLICE_ENCODING_DIRECTION,
+    SLICE_ENCODING_DIRECTION_TO_DIM,
     create_bids_slice_timing_from_coordinate,
     create_slice_time_coordinate_from_bids,
 )
@@ -51,6 +53,8 @@ __all__ = [
     "FUSIBIDSMetadata",
     "validate_metadata",
     # Coordinates
+    "SLICE_ENCODING_DIRECTION_TO_DIM",
+    "DIM_TO_SLICE_ENCODING_DIRECTION",
     "create_bids_slice_timing_from_coordinate",
     "create_slice_time_coordinate_from_bids",
 ]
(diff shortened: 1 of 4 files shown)


sdiebolt:
@FelipeCybis Good to review!

Issue #58, opened 2026-04-28
Atlas registration via another atlas space: Iconeus case


Anatole94:
Given a fUSI data and an affine transform for registration on the atlas but in other spaces, how to handle it?

In the case of Iconeus data, we have:

- a fUSI data .scan
- an affine transform registering the fUSI data in the Iconeus lab space to the atlas (reference) in the Iconeus atlas space

Anatole94:
[load_Iconeus_bps_data.ipynb](https://github.com/user-attachments/files/27165598/load_Iconeus_bps_data.ipynb)

Proposed solution which doesn't work:
**I construct the transform chain from fUSI physical to atlas space CCFv3 via the BPS (fUSI physical -> BPS -> CCFv3 and input it in the resample function), small jupyter notebook attached**

I provide:
- iconeus_lab_to_confusius_lab (Iconeus lab -> ConfUSIus lab) correcting for metric to millimetric and z-flipping
```
# Iconeus lab -> ConfUSIus lab
# Iconeus lab uses metric coordinates and Confusion lab uses millimetric coordinates, so we need to scale the transformation
scale_factor = 1000
iconeus_lab_to_confusius_lab = np.diag([scale_factor, scale_factor, scale_factor, 1.0])
# BrainToLab comes from fuslab with z-flipped data, so we need to unflip it to get the correct transformation
iconeus_lab_to_confusius_lab[2, 2] *= -1
```
- brain_to_lab (Brain -> Iconeus lab) from the bps file obtained after the registration of the fUSI to a reference in brain space
```
# Brain -> Iconeus lab
with File(bps_path, "r") as f:
    brain_to_lab = f["BrainToLab"][:]
```
- atlas_voxels_to_brain (Atlas voxel -> Brain) from the sform stored in the pyfUS fetched atlas
```
# Atlas voxel -> Brain
atlas_voxels_to_brain = loadmat(atlas_voxel_to_brain_iconeus_path)["affines"]
```
from PyfUS
```
_, atlas = fetch_atlas_allen()
scipy.io.savemat(atlas_voxel_to_brain_iconeus_path, atlas.affines)
```
- atlas_voxels_to_ccfv3 (Atlas voxel -> CCFv3) from the ConfUSIus atlas coordinates
```
# Atlas voxel -> CCFv3
atlas_voxels_to_ccfv3 = np.array([
    [0.025, 0, 0, 0 ],
    [0, 0.025, 0, 0],
[...]

PR #75, opened 2026-04-30
feat: add `.bps` support

### LINKED ISSUE #74: Affines in non consolidated scans are not carried through to the consolidated scan
## Summary

`consolidate_poses` rebuilds `attrs["affines"]` as a single-key dict, silently dropping every affine other than the one it consolidated.

## Reproduction

```python
```python
import numpy as np
import confusius as cf
from confusius.multipose import consolidate_poses

da = cf.load("multipose.scan")  # 3Dscan: physical_to_lab is (npose, 4, 4)

# Attach a second per-pose affine alongside physical_to_lab.
da.attrs["affines"]["physical_to_other"] = (
    np.linalg.inv(np.eye(4)) @ da.attrs["affines"]["physical_to_lab"]
)
print(da.attrs["affines"].keys())
# dict_keys(['physical_to_lab', 'physical_to_other'])

consolidated = consolidate_poses(da)
print(consolidated.attrs["affines"].keys())
# dict_keys(['physical_to_lab'])
```

## Expected

`consolidated.attrs["affines"]` retains every affine present on the input. Per-pose affines linearly linked to the main affine (e.g. physical_to_other[p] = inv(link) @ physical_to_lab[p]) should be consolidated alongside it, so the same chain holds against the consolidated physical_to_lab.

## Actual

Only the main physical_to_lab survives; physical_to_other (and any other linked per-pose affine) is silently dropped, breaking downstream code that expected to compose with some other space after consolidation.

FelipeCybis:
## Summary

Adds BPS (Iconeus Brain Positioning System) sidecar support to `load_scan` so that fUSI scans can carry the Iconeus brain frame alongside the existing lab frame.

Inspired by the discussion in #58, where the proper `BrainToLab` → ConfUSIus-lab change of basis was worked out. This PR turns that recipe into a supported feature.

## Changes

- **`load_scan(..., bps_path=...)`**: optional `bps_path` argument loads the `.bps` HDF5 sidecar and stores a `physical_to_brain` affine in `da.attrs["affines"]`. The affine maps ConfUSIus physical `(z, y, x, 1)` mm coordinates to Iconeus brain coordinates, and is broadcast across poses for `3Dscan` / `4Dscan` files. Brain-side units are intentionally not converted, the BPS format does not declare them.
- **`load_bps` API**: exposed via `confusius.io` so users can construct the `brain_to_lab` affine independently of `load_scan`.
- **`consolidate_poses` fix**: previously rebuilt `attrs["affines"]` as a single-key dict, silently dropping `physical_to_brain` (and any other linked per-pose affine) on consolidation. Now propagates linked per-pose affines through consolidation as `L @ consolidated_main`, where `L` is the constant left-link recovered from pose 0; raises `ValueError` if a per-pose affine in `attrs["affines"]` is not a constant left-link of the main affine. Closes #74.
- **User guide**: new "Loading a `.bps` File" subsection in `docs/user-guide/io.md`.

## Test plan

- [x] `tests/unit/test_io/test_scan.py::TestLoadScanWithBPS` — 6 tests covering
      attribute presence, affine shape (2D / 3D / 4D), analytic origin-point
      correctness check, and `bps_path` validation.
- [x] `tests/unit/test_multipose/test_consolidate.py::test_physical_to_brain_consolidated_with_bps`
      — verifies `consolidated_physical_to_brain == inv(brain_to_lab) @ consolidated_physical_to_lab`.
- [x] `tests/unit/test_multipose/test_consolidate.py::test_unlinked_extra_per_pose_affine_raises`
      — verifies the new `ValueError` guard.
- [x] Full unit suite passes (`uv run pytest tests/unit`).

### DIFF
--- a/docs/user-guide/io.md
+++ b/docs/user-guide/io.md
@@ -344,9 +344,9 @@ scope or call [`.compute()`][xarray.DataArray.compute] before discarding it.
 Provenance metadata from the file is stored in `da.attrs`: `scan_mode`, `subject`,
 `session`, `scan`, `project`, `date`, `neuroscan_version`, and `machine_sn`.
 
-#### Loading a `.bps` File
+#### Loading a BPS File
 
-Iconeus' `.bps` files are HDF5 containers produced by Iconeus' Brain Positioning System.
+Iconeus' BPS files are HDF5 containers produced by Iconeus' Brain Positioning System.
 They contain an affine matrix that maps Iconeus brain coordinates `(x_brain, y_brain,
 z_brain, 1)` to Iconeus lab coordinates `(x_lab, y_lab, z_lab, 1)` in meters. Iconeus'
 lab space axes are aligned with the probe: `x_lab = lateral`, `y_lab = elevation`,
@@ -354,7 +354,7 @@ lab space axes are aligned with the probe: `x_lab = lateral`, `y_lab = elevation
 `(z_lab, y_lab, x_lab)` (elevation, depth, lateral) in millimeters, matching the
 convention used throughout the rest of the package.
 
-The affine matrix in the `.bps` file can be loaded with [`confusius.io.load_bps`][confusius.io.load_bps].
+The affine matrix in the BPS file can be loaded with [`confusius.io.load_bps`][confusius.io.load_bps].
 
 ```python
 import confusius as cf
@@ -376,8 +376,8 @@ physical_to_lab = da.attrs["affines"]["physical_to_lab"]
 physical_to_brain = np.linalg.inv(bps) @ physical_to_lab
 ```
 
-In fact, if you pass the `.bps` file using the `bps_path` argument when loading a
-`.scan` file with [`confusius.load`][confusius.load], the `physical_to_brain` affine
+In fact, if you pass the BPS file using the `bps_path` argument when loading a
+SCAN file with [`confusius.load`][confusius.load], the `physical_to_brain` affine
 will be computed automatically and stored in the resulting DataArray's attributes
 `affines` alongside the `physical_to_lab` affines:
 
--- a/docs/user-guide/io.md
+++ b/docs/user-guide/io.md
@@ -363,7 +363,8 @@ bps = cf.io.load_bps("sub-01_task-awake_pwd.bps")
 ```
 
 A `physical_to_brain` matrix between Iconeus' brain space and ConfUSIus' physical space
-can be established by using the `physical_to_lab` affines from the SCAN file:
+can be established and added to the DataArray's attributes by using the
+`physical_to_lab` affines from the SCAN file:
 
 ```python
 import confusius as cf
@@ -374,6 +375,7 @@ bps = cf.io.load_bps("sub-01_task-awake_pwd.bps")
 
 physical_to_lab = da.attrs["affines"]["physical_to_lab"]
 physical_to_brain = np.linalg.inv(bps) @ physical_to_lab
+da.attrs["affines"]["physical_to_brain"] = physical_to_brain
 ```
 
 In fact, if you pass the BPS file using the `bps_path` argument when loading a
--- a/docs/user-guide/io.md
+++ b/docs/user-guide/io.md
@@ -344,6 +344,30 @@ scope or call [`.compute()`][xarray.DataArray.compute] before discarding it.
 Provenance metadata from the file is stored in `da.attrs`: `scan_mode`, `subject`,
 `session`, `scan`, `project`, `date`, `neuroscan_version`, and `machine_sn`.
 
+#### Loading a `.bps` File
+
+If you have an Iconeus `.bps` sidecar (a Brain Positioning System file produced
+alongside the SCAN), pass it via the `bps_path` argument so the resulting DataArray
+carries an extra `physical_to_brain` affine in `da.attrs["affines"]`:
+
+```python
+import confusius as cf
+
+da = cf.load(
+    "sub-01_task-awake_pwd.scan",
+    bps_path="sub-01_task-awake_pwd.bps",
+)
+
+physical_to_brain = da.attrs["affines"]["physical_to_brain"]
+```
+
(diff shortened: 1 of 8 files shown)


FelipeCybis:
@sdiebolt good to review :)

docs/user-guide/io.md
```diff
@@ -344,6 +344,59 @@ scope or call [`.compute()`][xarray.DataArray.compute] before discarding it.
...
+
+```python
+import confusius as cf
+import numpy as np
+
+da = cf.load("sub-01_task-awake_pwd.scan")
+bps = cf.io.load_bps("sub-01_task-awake_pwd.bps")
+
+physical_to_lab = da.attrs["affines"]["physical_to_lab"]
+physical_to_brain = np.linalg.inv(bps) @ physical_to_lab
```
sdiebolt, line 376: Since we're always using the lab to brain transform, would it make sense for `load_bps` to actually return lab->brain instead of brain->lab? That way users don't have to actually compute the inverse themselves?

docs/user-guide/io.md
```diff
@@ -344,6 +344,59 @@ scope or call [`.compute()`][xarray.DataArray.compute] before discarding it.
 Provenance metadata from the file is stored in `da.attrs`: `scan_mode`, `subject`,
 `session`, `scan`, `project`, `date`, `neuroscan_version`, and `machine_sn`.
 
+#### Loading a `.bps` File
+
+Iconeus' `.bps` files are HDF5 containers produced by Iconeus' Brain Positioning System.
+They contain an affine matrix that maps Iconeus brain coordinates `(x_brain, y_brain,
+z_brain, 1)` to Iconeus lab coordinates `(x_lab, y_lab, z_lab, 1)` in meters. Iconeus'
+lab space axes are aligned with the probe: `x_lab = lateral`, `y_lab = elevation`,
+`z_lab = axial`. We chose to re-express this space as **ConfUSIus-ordered** lab space
```
sdiebolt: Lab axes aren't always aligned with the probe I think? It's only at the beginning of the acquisition, and then it can be different depending on how the probe moved.

docs/user-guide/io.md
sdiebolt: Also here we could also show how to assign this new `physical_to_brain` affine to the DataArray's attributes.

docs/user-guide/io.md
```diff
@@ -344,6 +344,59 @@ scope or call [`.compute()`][xarray.DataArray.compute] before discarding it.
 Provenance metadata from the file is stored in `da.attrs`: `scan_mode`, `subject`,
 `session`, `scan`, `project`, `date`, `neuroscan_version`, and `machine_sn`.
 
+#### Loading a `.bps` File
+
+Iconeus' `.bps` files are HDF5 containers produced by Iconeus' Brain Positioning System.
```
sdiebolt: To refer to file formats, I've been following Google's style guide for a long time: I think we should stick to it since most of the package should already follow that convention: https://developers.google.com/style/filenames#file-type-names

Would that be OK with you? If so, we should refer to BPS and SCAN files, not `.bps`/`.scan`.

src/confusius/io/scan.py
```diff
@@ -181,8 +181,65 @@ def _build_physical_to_lab(
     return physical_to_lab
 
 
+def load_bps(bps_path: str | Path) -> npt.NDArray[np.float64]:
+    """Load a `.bps` file and return a `brain_to_lab` affine in ConfUSIus lab space.
```
sdiebolt:
```suggestion
    """Load a BPS file and return an affine from Iconeus' brain space to ConfUSIus lab space.
```

src/confusius/io/scan.py
```diff
@@ -181,8 +181,65 @@ def _build_physical_to_lab(
...
+
+    BPS files are HDF5 sidecars produced by Iconeus' brain positioning system. They
+    store a `BrainToLab` affine that maps Iconeus brain coordinates `(x_brain, y_brain,
+    z_brain, 1)` to Iconeus lab coordinates `(x_lab, y_lab, z_lab, 1)` in meters.
+    Iconeus' lab axes are aligned with the probe: `x_lab = lateral`, `y_lab =
+    elevation`, `z_lab = axial`.
```
sdiebolt: Same thing as above: this is only true at the beginning of an acquisition.

src/confusius/io/scan.py
```diff
@@ -181,8 +181,65 @@ def _build_physical_to_lab(
...
+    BPS format and are therefore not converted).
+
+    The change of basis from ConfUSIus-ordered millimetre lab coordinates to
+    Iconeus-ordered metre lab coordinates is
+
+    ```
+    confusius_lab_to_iconeus_lab = mm_to_m @ P
+    ```
+
+    where `P = _PHYSICAL_TO_PROBE_PERMUTATION` permutes the axes from ConfUSIus
```
sdiebolt: If we start mentioning `_PHYSICAL_TO_PROBE_PERMUTATION`, we should make it public?

FelipeCybis: That is probably a good idea.

docs/user-guide/io.md
FelipeCybis: Got it!

docs/user-guide/io.md
FelipeCybis: indeed, what I wanted to say is that it is linked to the probe axes through the `physical_to_lab` affine. I will rephrase it.

docs/user-guide/io.md
FelipeCybis: > Since we're always using the lab to brain transform, would it make sense for `load_bps` to actually return lab->brain instead of brain->lab? That way users don't have to actually compute the inverse themselves?

I am not sure, I think, if we can, we should stick to what the affine actually means at source. IMHO, I think that making the user think a small bit on the inverse of the transform and why is that is actually a good thing.
We are providing the automatic path anyways when loading the bps together with the scan.

I will add the assignment.

sdiebolt (approved):
LGTM! 🚀

PR #93, opened 2026-05-05
chore: prepare v0.2.0 release

sdiebolt:
## Summary
- finalize version metadata for `v0.2.0`
- convert the changelog entry from development to released form
- update release citation metadata and docs references

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -32,5 +32,5 @@ keywords:
   - Brain connectivity
   - Python
 license: BSD-3-Clause
-version: 0.0.1-a26
-date-released: '2026-05-01'
+version: 0.2.0
+date-released: '2026-05-05'
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ and tutorials.
 
 If you use ConfUSIus in your research, please cite it using the following reference:
 
-> Le Meur-Diebolt, S., & Cybis Pereira, F. (2026). ConfUSIus (v0.0.1-a26). Zenodo.
+> Le Meur-Diebolt, S., & Cybis Pereira, F. (2026). ConfUSIus (v0.2.0). Zenodo.
 > https://doi.org/10.5281/zenodo.18611124
 
 Or in BibTeX format:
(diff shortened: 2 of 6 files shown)


FelipeCybis:
@sdiebolt i think we should remove the uv.lock file? See https://github.com/astral-sh/uv/issues/10730

sdiebolt:
> @sdiebolt i think we should remove the uv.lock file? See [astral-sh/uv#10730](https://github.com/astral-sh/uv/issues/10730)

I think I'm more on Charlie Marsh's side? 

https://github.com/astral-sh/uv/issues/10730#issuecomment-2599758198

Given that we don't test against specific package versions, do you see an advantage in not having the `uv.lock` file? I think it gives more reproducibility to our dev environment, as it ensures every dev uses the same local versions? 

Also, I think this gives a few good views on lockfiles: https://blog.rust-lang.org/2023/08/29/committing-lockfiles/
I'd prefer keeping the lockfile and maybe implementing some CI job in the future to keep it up to date with the latest package versions?

FelipeCybis:
Ok, I see, that is a dev-only think in libraries. I do think that a CI for updating it is indeed needed, then. I will check how people do it.

sdiebolt:
> I do think that a CI for updating it is indeed needed, then. I will check how people do it.

Should we rather put that in a separate PR after the 0.2.0 release? I think it might go well in an independent infrastructure PR, since it doesn't affect the package itself (only the development environment, i.e. only the two of us for now).

FelipeCybis:
> > I do think that a CI for updating it is indeed needed, then. I will check how people do it.
> 
> 
> 
> Should we rather put that in a separate PR after the 0.2.0 release? I think it might go well in an independent infrastructure PR, since it doesn't affect the package itself (only the development environment, i.e. only the two of us for now).
> 
> 

Yes yes, I was thinking on a different PR!

FelipeCybis (approved):

sdiebolt:
> Yes yes, I was thinking on a different PR!

I leave that up to you then, I remember you had created some pretty robust CI jobs for that kind for release/docs/maintenance work in a previous life that was super useful :stuck_out_tongue:

PR #132, opened 2026-05-13
  fix: consolidate [tool.uv] and fix win32 dependency resolution

EstebanRivera08:
## Description

  - Remove `platform_machine == 'AMD64'` from `required-environments`: uv 0.6.9
    falsely rejects `pywin32>=303` under that constraint despite win_amd64 wheels
    existing, blocking lock file resolution.
  - Add `override-dependencies` to exclude `cartopy` and `napari-cartopy` on
    Windows: both pull C extensions unavailable on win32 (pulled  by `xarray[complete]` and `napari[all]`).
  - Sync `uv.lock` to reflect the new resolution.


sdiebolt:
Hey @EstebanRivera08, thanks for contributing! 

I'm curious about cartopy not working on Windows. Our test suite is running on Windows and didn't catch any problem, and I think @FelipeCybis is also running Windows. It also seems weird to me that Xarray and napari, which are very well known packages, might not be working on Windows (`napari[all]` is the recommended install for all users, and `xarray[complete]` is pretty common). Would you have  more information on the error you get during install?

Also I see your comment mentioning uv 0.6.9, which is already a pretty old version (from early 2025). Could you try to upgrade uv to the latest version (0.11.14) and see if that fixes this problem?

EstebanRivera08:
Hi!

Indeed, when I first downloaded and ran it on my desktop Windows machine some days ago, everything worked without errors. But today, when trying it on my laptop, I ran into the following error (which disappeared once I avoided pulling cartopy, as mentioned earlier):

uv run confusius
Using CPython 3.14.0a6
Creating virtual environment at: .venv
      Built confusius @ file:///C:/Users/deyve/Documents/test/confusius
  × Failed to build `cartopy==0.25.0`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_wheel` failed (exit code: 1)
...
ImportError: DLL load failed while importing _multiarray_umath: The specified procedure could not be found.
...

      Importing the numpy C-extensions failed. This error can happen for
      many reasons, often due to issues with your setup or how NumPy was
      installed.

      We have compiled some common reasons and troubleshooting tips at:

          https://numpy.org/devdocs/user/troubleshooting-importerror.html

      Please carefully study the information and documentation linked above.
      This is unlikely to be a NumPy issue but will be caused by a bad install
      or environment on your machine.

      Original error was: DLL load failed while importing _multiarray_umath: The specified procedure could not be found.


      hint: This usually indicates a problem with the package or the build environment.
  help: `cartopy` (v0.25.0) was included because `confusius` (v0.3.0.dev0) depends on `xarray[complete]` (v2026.2.0) which depends on `cartopy`.

I just realized that after running uv python pin 3.13, everything works correctly. So I think that is mostly something related to instability with python 3.14 (at least in my machine)? :)

sdiebolt:
@EstebanRivera08 from the error message, I see you're using a pre-release Python 3.14 (CPython 3.14.0a6)! That might be why the cartopy install errors (maybe trying to get built from source?). You should be able to install a recent 3.14 version using uv (`uv python install 3.14`) and that should fix your problem! ConfUSIus passes our tests on Windows + Python 3.14 in CI. If that doesn't fix it, do tell us, we'll investigate!

FelipeCybis:
Hi @EstebanRivera08! Indeed, `cartopy` do not have pre-built wheels for Python 3.14. So I think the errors you are seeing come from not having the necessary build tools to build `cartopy` yourself (even though that is a numpy error that is surfacing). In Windows these tools would be the MSVC Build Tools for C++.

PR #136, opened 2026-05-14
fix(ci): handle empty amend in docs deploy when content unchanged

### LINKED ISSUE #135: Docs deploy fails with empty amend when content unchanged
## Problem

The `Deploy dev docs` step in `.github/workflows/docs.yml` (and the analogous `Deploy release docs` step) fails when documentation content has not changed since the previous deploy.

## Symptom

Workflow log:

```
Build started
No issues found
Build finished in 15.42s
You asked to amend the most recent commit, but doing so would make
it empty. You can repeat your command with --allow-empty, or you can
remove the commit entirely with "git reset HEAD^".
On branch main
Your branch is ahead of 'origin/main' by 1 commit.
  (use "git push" to publish your local commits)

No changes
Error: Process completed with exit code 1.
```

## Root cause

The deploy step does:

1. Save `site/examples/index.html` to `/tmp/`.
2. Run `mike deploy ... --update-aliases dev latest`, which produces a new commit on the docs branch.
3. Overwrite `dev/examples/index.html` with the saved `/tmp` copy.
4. `git add dev/examples/index.html`.
5. `git diff --cached --quiet || git commit --amend --no-edit`.

When the only file `mike deploy` actually changed against the gh-pages parent is `dev/examples/index.html`, step 3 reverts that change. Step 4 stages a diff (mike's version → restored version), so step 5 attempts the amend. The amended commit's net delta against the parent is empty, and `git commit --amend` aborts with the "would make it empty" error, failing the job.

The check `git diff --cached --quiet` only asks "are there staged changes?" — not "will the amended commit be non-empty vs the parent?".

## Fix

Tracked in PR #136 on branch `fix/docs-deploy-empty-amend`. Approach:

- If `git commit --amend` fails with "would make it empty", drop mike's commit (`git reset --hard HEAD~1`) — semantically correct, since there is nothing to deploy.
- Push only when `HEAD` differs from `origin/main`.
- Other amend errors still propagate.

Applied to both the `Deploy dev docs` and `Deploy release docs` steps.

FelipeCybis:
Closes #135.

## Summary
- In both `Deploy dev docs` and `Deploy release docs`, detect the "amend would make it empty" failure from `git commit --amend` and drop mike's commit instead of erroring out.
- Push only when `HEAD` differs from `origin/main`, so a dropped commit produces a clean no-op deploy.
- Other amend errors still propagate (`exit 1`).

## Test plan
- [ ] Trigger a docs build with no doc-content changes since the last `main` deploy and verify the workflow succeeds with `Amend would be empty (no effective doc changes); dropping commit` followed by `No changes vs origin/main; skipping push`.
- [ ] Trigger a docs build with a real doc change and verify the deploy completes and pushes as before.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -238,9 +238,23 @@ jobs:
           uv run --project ${{ github.workspace }} mike deploy --config-file ${{ github.workspace }}/zensical.toml --branch main --update-aliases dev latest
           cp /tmp/confusius-examples-index.html dev/examples/index.html
           git add dev/examples/index.html
-          git diff --cached --quiet || git commit --amend --no-edit
+          if ! git diff --cached --quiet; then
+            if ! git commit --amend --no-edit 2>/tmp/amend.err; then
+              if grep -q "make it empty" /tmp/amend.err; then
+                echo "Amend would be empty (no effective doc changes); dropping commit"
+                git reset --hard HEAD~1
+              else
+                cat /tmp/amend.err >&2
+                exit 1
+              fi
+            fi
+          fi
           uv run --project ${{ github.workspace }} mike list --config-file ${{ github.workspace }}/zensical.toml --branch main | grep -q "^stable " || uv run --project ${{ github.workspace }} mike set-default --config-file ${{ github.workspace }}/zensical.toml --branch main latest
-          git push origin main
+          if git diff --quiet origin/main HEAD; then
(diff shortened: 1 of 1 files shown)


sdiebolt:
Unfortunately this fixes a temporary hack that I created but now need to remove! I'm currently working on making the deploy work without having to clone `confusius-docs`, so this whole amending-the-mike-commit step will go away shortly. :melting_face:

PR #139, opened 2026-05-14
feat(registration)!: return RegistrationDiagnostics from register_volume

### LINKED ISSUE #138: Feature request: returning the registration metrics in `register_volume`

There is currently no way to recover the metrics used during registration. This is useful for a before/after visualization.

Image

FelipeCybis:
## Description

Closes #138.

`register_volume` had no way to return the optimization metrics, so the before/after convergence visualisations users want had to be re-instrumented by hand. This PR adds a `RegistrationDiagnostics` dataclass that travels alongside every registration call.

### What changes

- **New dataclass** `confusius.registration.RegistrationDiagnostics`  (`metric`, `metric_values`, `final_metric_value`, `n_iterations`,  `stop_condition`).
- `register_volume` always attaches a `sitkIterationEvent` observer that records `GetMetricValue()` per iteration, independently of `show_progress`, and returns it as a third tuple element.
- `register_volumewise` aggregates per-frame diagnostics under `attrs["registration_diagnostics"]` and adds `final_metric_value` / `n_iterations` columns to `motion_params`.
- The `.fusi.register.to_volume` xarray accessor's return type follows.
- New docs example `Registering two acquisitions` walks through registering two Nunez-Elizalde 2022 sessions, with confusius `plot_volume` overlays for before/after composites and a diagnostics-driven convergence plot. Two callouts cover argument sensitivity and the live `show_progress=True` view.

### Breaking change

`register_volume` now returns `(registered, transform, diagnostics)` instead of `(registered, transform)`. Existing callers need an extra unpack target (or `_`).

## Checklist

- [x] Tests pass locally (`just t`) — 4 new `TestRegisterVolumeDiagnostics` tests
- [x] New public API is documented (`RegistrationDiagnostics` docstring, updated `register_volume` / `register_volumewise` / accessor docstrings)
- [x] Changelog updated (new `Breaking changes`, `Enhancements`, and `Documentation` bullets, all linked to #138)


FelipeCybis:
@sdiebolt should be good to be reviewed

src/confusius/registration/volumewise.py
```diff
@@ -194,9 +199,11 @@ def register_volumewise(
     arr = data_moved.values
     output = np.zeros_like(arr)
     affines: list[npt.NDArray[np.floating] | None] = []
-    for t, (registered_da, frame_affine) in enumerate(results):
+    diagnostics: list[RegistrationDiagnostics] = []
+    for t, (registered_da, frame_affine, frame_diag) in enumerate(results):
         output[t] = registered_da.values
         affines.append(frame_affine)
+        diagnostics.append(frame_diag)
```
sdiebolt: Wouldn't that potentially create a lot of data? We should probably have an argument to allow skipping it, as it wouldn't be often useful.

docs/examples/registration/register_volume_two_acquisitions.py
```diff
@@ -0,0 +1,207 @@
...
+    show_titles=False,
+)
+plotter.add_volume(
+    moving_on_fixed,
+    cmap="Blues",
+    alpha=0.5,
+    match_coordinates=False,
+    show_colorbar=False,
+    show_titles=False,
+)
```
sdiebolt: We should introduce a composite plotter that creates an RGB image (red/cyan) instead of plotting using transparency. Maybe in a separate PR that we merge before this one?

sdiebolt (commented):
First quick pass

docs/changelog.md
```diff
@@ -10,8 +10,23 @@ icon: lucide/history
...
+
 ### :sparkles: Enhancements
 
+- `register_volume` now also returns a
+  [`RegistrationDiagnostics`][confusius.registration.RegistrationDiagnostics] dataclass
+  with the per-iteration metric values, final metric value, iteration count, optimizer
+  stop condition, and the metric name. `register_volumewise` propagates the per-frame
+  diagnostics list under `attrs["registration_diagnostics"]` and adds
+  `final_metric_value` and `n_iterations` columns to `motion_params`
+  ([#139](https://github.com/confusius-tools/confusius/pull/139)).
```
sdiebolt: Do we need to repeat that if it was already mentioned in the breaking changes? I'd leave only one, maybe this one because it's more detailed, and move it to the breaking changes section.

docs/examples/registration/register_volume_two_acquisitions.py
FelipeCybis: Tracking PR here #145

src/confusius/registration/volumewise.py
FelipeCybis: Yes, it would potentially make an array of a few hundred values per frame (potentially a few thousands). `keep_diagnostics` should be set to `True` for this to happen (fixed in https://github.com/confusius-tools/confusius/pull/139/commits/a5edbd20b35bc7402d313ce3a7751e14164129af)

docs/changelog.md
FelipeCybis: Indeed. Done here https://github.com/confusius-tools/confusius/pull/139/commits/b6edf9573688a1961bec36ed31a65c331cbbe87c

FelipeCybis:
@sdiebolt all good for another round of review, now with `plot_composite` merged :P

sdiebolt (approved):
LGTM! :rocket:

Issue #151, opened 2026-05-19
Add function to validate a ConfUSIus-expected Data Array


FelipeCybis:
We should have a validating function to enforce metadatas, data structures, coordinates, dimensions, etc. that we expect to use across the library. We do have `validate_iq` as inspiration.

FelipeCybis:
I am thinking on this. It would be useful to brainstorm a bit on what should be validated to a DataArray to be considered a fUSI DataArray. It seems a good idea to be able to opt in or out of the validations

- At least 3D spatial dimensions. `SPATIAL_DIMS = ("z", "y", "x")` (should it be possible to opt-out?)
- Opt-in to require `time` (preprocessing/analysis functions).
- Require `time` (if exists) to be monotonic-increasing? (not sure if this changes something in the analysis functions)
- Require `dims` to be strings? (I saw stringification of dims in the code, but not sure this would be needed)
- Require `coords` to have `voxdim` on `SPATIAL_DIMS`.
- Require `SPATIAL_DIMS` to have strictly monotonic-increasing coordinates (#150)

sdiebolt:
Here's my 2 cents on the validation function (e.g. named `validate_fusi_dataaarray`, and we should rename `validate_iq` -> `validate_iq_dataarray` I think):

1. **Type**: must be a DataArray.
2. **Dimensions**: warn if the order is not `(time, pose, z, y, x)`, allow at least 2D (e.g., `(y, x)`, I think most function do well with `(time, y, x)` 2D+t data).
3. **Coordinates**: all dimensions have associated coordinates, and that coordinates are monotonically increasing when numeric (with an arg to disable?). We could have args to enforce regular spacing, allow multiple poses, require a time dimension, etc.
4. **Metadata**: we could warn about nice metadata to have (`voxdim`, `volume_acquisition_reference`/`_duration`, etc.)

Issue #150, opened 2026-05-19
Should axis coordinates be monotonic-increasing?


FelipeCybis:
#146 raised a point that was not completely addressed, whether we should have an understanding of the direction of the axis coordinates.

Indeed, dimensions with coordinates that are not monotonic-increasing (monotonic-decreasing or scrambled coordinates) may cause different problems throughout ConfUSIus:

- Plotting functions don't work well with data that are not monotonically ordered. This is not exactly a problem for monotonic-decreasing coordinates, but we currently sort the data inside plotting functions before actually passing them to napari, matplotlib, etc.
- Resampling/registration functions are based on `SimpleITK`. A SimpleITK image can be constructed with an `origin` and a (positive) `spacing`. Today, data with monotonic-decreasing coordinates create a negative spacing and SimpleITK raises an error.
- The negative spacing of monotonic-decreasing coordinates is also a problem for `smooth_volume`. A negative spacing for the Gaussian function is a no-op for the given dimension.

After a quick discussion with @sdiebolt, we have some possible solutions (not mutually exclusive):
- [ ] Force monotonic-increasing coordinates via a `validate_fusi` function. Meaning that this is what we will expect in the different functions in `confusius`.
    - This could come with an argument on loading functions to automatically sort the coordinates (defaulting to `True` for enforcing it or not).
- [ ] Silently (or with a warning) sort the data coordinates when needed (cf above).
    - This solution has some problems as it creates some ambiguity on the direction of the coordinates of output data in resampling/registration function.
- [ ] Error if data do not have monotonic-increasing coordinates whenever is needed. Point user on how to sort the data coordinates.

sdiebolt:
After thinking about it a bit more, I lean toward the third option: raise an error via `validate_fusi` when DataArrays don't have monotonic-increasing coordinates in functions that require them. I think such recordings will be very rare (mainly from weird manipulation of loaded DataArrays, or incorrectly constructed DataArrays from custom file formats, like the example from #146), and we should make users aware that their DataArray is weird rather than silently handling it behind their backs. We should thus introduce a helper function to reorder DataArrays to have monotonic-increasing coordinates, so that users can fix their recordings!

FelipeCybis:
[...]

PR #152, opened 2026-05-19
refactor: handle monotonic-decreasing coordinates

FelipeCybis:
This is a WIP for closing #150 and #151 

Currently, this moves the plotting helper function to sort coordinates into `confusius._utils` and refactor plotting and resampling/registration functions to use it and force monotonic-increasing coordinates in the data.

`smooth_volume` uses abs(spacing) as this is the expected result even if data coordinates are monotonic-decreasing.

## Checklist

- [ ] Tests pass locally (`just t`)
- [ ] Pre-commit hooks pass (`just pc`)
- [ ] Docstring updates after deciding how to handle non-monotonic-increasing coordinates
- [ ] Changelog updated if user-facing
- [ ] Add information about what is decided here in the User Guide

### DIFF
--- a/src/confusius/_napari/_io/_readers.py
+++ b/src/confusius/_napari/_io/_readers.py
@@ -16,7 +16,10 @@ from typing import TYPE_CHECKING, Any, Callable
 from napari.layers.utils.layer_utils import calc_data_range
 from napari.utils.notifications import show_warning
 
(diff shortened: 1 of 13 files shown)


FelipeCybis:
Handled more broadly with #153

PR #153, opened 2026-05-19
feat(validation): add generic dataarray validators

### LINKED ISSUE #150: Should axis coordinates be monotonic-increasing?

### LINKED ISSUE #151: Add function to validate a ConfUSIus-expected Data Array

sdiebolt:
## Summary
- add `validate_fusi_dataarray` for shared ConfUSIus DataArray validation
- refactor IQ validation to build on the generic validator
- rename `validate_iq` to `validate_iq_dataarray`
- update docs and validation tests

## Testing
- `uv run pytest tests/unit/test_validation/test_fusi_dataarray.py tests/unit/test_validation/test_validation_iq.py tests/unit/test_validation/test_coordinates.py`
- pre-commit hooks via `git commit`

Closes #151
Closes #150


sdiebolt:
@FelipeCybis Ready for a first look. For now I've used the validation function in the registration and spatial modules. Maybe there're other places where we could use it?

src/confusius/spatial/smooth.py
```diff
@@ -122,6 +124,12 @@ def smooth_volume(
 
     >>> smoothed = smooth_volume(data, fwhm=0.3, ensure_finite=True)
     """
+    try:
+        validate_fusi_dataarray(data, minimum_spatial_dims=1)
+    except ValueError as exc:
+        if "Missing required coordinate for dimension" not in str(exc):
+            raise
+
```
FelipeCybis, line 128: Shouldn't we allow dimensions without coordinates through a flag to validate_fusi_dataarray instead?

sdiebolt: Hmm... I wondered about that last week but finally decided that all dimensions should have coordinates. It's obvious for core dims, but even for non-core dims I don't see any advantage in not having coordinates (e.g., regions have region IDs, components have component indices, etc). Do you see any example where having coordinates  would be a pain?

FelipeCybis: Hmm I don't think that is the issue here, I just don't think we should make an exception just for this method. If we want to allow dimensions without coordinates for smoothing, I think this should be done clearly through validation.

But about forcing coordinates for all dimensions... seems a bit harsh. I find useful to concatenate scans just to plot them using confusius' plotting tools, for example. I don't see this as an anti pattern for custom dimensions.

sdiebolt: Makes sense. Would enforcing coordinates on core dimensions but not necessarily on extra dimension be OK with you?

FelipeCybis: Yes, I think this is a good compromise! What do you see the correct logic for smoothing then? Default validation (no coords on core dims = error)?

sdiebolt: Yup, that would make sense, wouldn't it? I guess we should generally constrain users to use coords everywhere it makes sense, even when we're a bit too strict (e.g. smoothing could require having coords only on smoothed dims, but I guess it doesn't hurt to remind the user to user coords on all core dims?).

sdiebolt: 204a112 does this, I'll let you review!

FelipeCybis (approved):
LGTM!

PR #159, opened 2026-05-21
test(xarray): add forwarding coverage for accessor wrappers

sdiebolt:
## Summary
Add focused monkeypatch-based unit tests for thin `xarray` accessor wrappers to improve wrapper coverage without duplicating implementation tests.

## What is covered
- `fusi.scale`
- `fusi.extract`
- `fusi.affine`
- `fusi.iq`
- `fusi.register`
- `fusi.connectivity.seed_map`
- `fusi.plot`

## Test strategy
For wrapper-heavy methods, tests assert that wrappers call the correct underlying function/class with forwarded arguments and return the wrapped result.

This follows the project testing guidance: no garbage-output assertions and reuse of shared fixtures from `tests/unit/conftest.py`.

### DIFF
--- a/tests/unit/test_xarray/test_accessors.py
+++ b/tests/unit/test_xarray/test_accessors.py
@@ -537,3 +537,24 @@ class TestAffineApplyMethod:
             result.coords["y"].values, da.coords["y"].values + 5.0
         )
         assert "x" not in result.coords
+
+    def test_unexpected_affine_shape_is_passed_through(self):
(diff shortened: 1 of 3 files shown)


sdiebolt:
@FelipeCybis ready to review! Should be a quick one, I'm basically just verifying that args passed to the accessors are indeed used by the wrapped functions.

FelipeCybis (approved):

PR #176, opened 2026-06-11
feat(napari): add events framework to napari plugin

### LINKED ISSUE #174: [Napari plugin] Add Events framework

I think look at [BIDS Events](https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/events.html) may be a good strategy. This gives us an Events structure directly.

In the BIDS Events, `onset` and `duration` are REQUIRED. Different events can be identified with `trial_type`.

I imagined several ways on looking at these events (not mutually exclusive):
- Being able to plot the events in a signals style matplotlib plot. Probably by painting the background of the plot instead of adding a new line plot.
- It could be cool to add a contour to the fUSI frame when the event is ON.
- Add some text to the text overlay showing when the event is ON or OFF.

The second useful step here is allowing the user to create a BIDS events file (.tsv) from napari. Sort of a temporal annotation feature.

FelipeCybis:
- Close #174.

## Summary
- add an **Events** panel to the napari sidebar for annotating *periods of time* (BIDS events: `onset`, `duration`, optional `trial_type`)
- annotate interactively with a Start/End workflow and S/E/Esc shortcuts (active only while the panel is open); reject zero/negative-duration events
- shade event bands on the Signals plot (colored per trial type) and name the active events in the time overlay, each toggleable
- load/save BIDS events `.tsv` files via `confusius.bids.read_events` / `write_events`
- represent events as a pandas DataFrame (`onset`/`duration`/`trial_type`), the same table the GLM already consumes, so a saved `.tsv` feeds straight into `make_first_level_design_matrix` and `FirstLevelModel.fit` with no conversion
- preserve every column of a loaded `.tsv` across a load → save round trip, including ones the plugin does not display
- document the panel in "Using the Plugin" (with a GIF of the Start/End workflow) and add a tip on reusing the saved `.tsv` in the GLM
- add unit tests for the events I/O, the `EventStore`, and the panel/plotter/ overlay integration

## Preview
plugin-events-create

## Checklist


FelipeCybis:
@sdiebolt this is a gift for whenever you are back at the office :)
Ready to review.

sdiebolt:
I'll review once I'm finished with #216! In the meantime, a few remarks:
- Should we maybe add a text overlay to show the name of the current event, e.g. next to the time overlay?
- We should update the take a tour feature to include this tab.

FelipeCybis:
> - Should we maybe add a text overlay to show the name of the current event, e.g. next to the time overlay?

That's already the case, whenever the time index is within an event period the name of the event appears on the right of the time overlay (if that is not the case there have been a regression and I should check it out)


> - We should update the take a tour feature to include this tab.

Indeed, I will do that.

Btw, we are adding two more panels to the widget, should we maybe take a moment to think if there is a better layout arrangement for the feature groups?

sdiebolt:
> That's already the case, whenever the time index is within an event period the name of the event appears on the right of the time overlay (if that is not the case there have been a regression and I should check it out)

Nice! My bad, I had only checked the GIF and was waiting on finishing the registration tab to create a new worktree to test it out. :stuck_out_tongue: 

> Btw, we are adding two more panels to the widget, should we maybe take a moment to think if there is a better layout arrangement for the feature groups?

I tested a few things a few weeks ago and couldn't find anything better... But I would be very interested if you have ideas!

sdiebolt (approved):
LGTM! Huge feature that will be be super useful for some users that already asked for it a few weeks ago!

PR #180, opened 2026-06-15
fix(napari): use slider axis for Signals x-axis without time dim

### LINKED ISSUE #179: [Napari plugin] Signals plot: x-axis uses step indices + "Time" label for data without a time dimension
### ConfUSIus version

Since v0.2.0

### Python version

any

### Operating system

WSL

### Installation

using uv

### Description

When a static 3D volume (no time dimension) is loaded into the napari plugin, the Signals plot draws the trace against step indices and labels the x-axis "Time", while the slider cursor moves in world coordinates (mm). The trace and the cursor therefore live on two different x-scales and never line up, and the x-axis dimension dropdown is empty.

Expected

- Trace plotted against the slider axis world coordinates (z, 0–4 mm), labeled Z (mm).
- Cursor tracks the trace end-to-end (far-right cursor == trace max == 4 mm).
- The x-axis dropdown offers the slider axis (z).

Actual

- Trace over step indices 0–40, labeled "Time" (there is no time dimension).
- Cursor only reaches ~4 (z world coords) — it stops a bit before the "5" tick while the trace extends to 40.
- The x-axis dropdown is empty/disabled.

Root cause

The x-axis dimension is resolved inconsistently:

- Cursor uses viewer.dims.point[idx] → world mm. ✅
- SignalPlotter._get_xaxis_coords / _get_xaxis_label default to the time dimension; with no time dim they fall back to np.arange(n) (indices) and "Time". ❌
- SignalPanel._get_available_xaxis_dims excludes all spatial axes, so the combo is empty for (z,y,x) data → _xaxis_dim stays None → the time fallback is used.

Regression

Introduced in f56f9a7. It changed the dropdown filter from non-displayed (slider) axes to exclude spatial axes:

```
-                if i not in displayed_dims and da.shape[i] > 1
+                if dim not in ALL_SPATIAL_DIMS and da.shape[i] > 1
```

This conflates "spatial" with "displayed" and breaks pure-spatial volumes, where a spatial axis is the slider. The same commit also added the else TIME_DIM fallback in the plotter.

Proposed fix

- Restore the original criterion in _get_available_xaxis_dims: offer the non-displayed (slider) axes with size > 1 (so (z,y,x) → [z], (time,z,y,x) → [time, z]).
- Resolve the x-axis dimension in _get_xaxis_coords / _get_xaxis_label via the existing _xaxis_dim_index (so trace, label, and cursor can't diverge) and drop the hard-coded "Time" in _render_imported_only.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#
# for issues.
import napari

from confusius._napari import ConfUSIusWidget
from confusius.datasets import fetch_nunez_elizalde_2022
from confusius.io import load
from confusius.plotting.napari import plot_napari

bids_root = fetch_nunez_elizalde_2022(
    subjects="CR022", sessions="20201007", tasks="__angio_only__"
)
pwd_file = (
    bids_root
    / "sub-CR022"
    / "ses-20201007"
    / "angio"
    / "sub-CR022_ses-20201007_pwd.nii.gz"
)

da = load(pwd_file).compute()

viewer = napari.Viewer()
viewer.window.add_dock_widget(ConfUSIusWidget(viewer), name="ConfUSIus")
_viewer, layer = plot_napari(da, viewer=viewer)
viewer.layers.selection.active = layer

napari.run()
```

### Additional output

We can see here that the combo box to select the x-axis is empty, even though we have the z dimension in the slider.
Also, the x-axis resolver resolved the x-axis label to "Time" and the x-axis ticks to the number of steps in this dimension. The x-axis cursor finds the world coordinates instead and there is an ugly mismatch.

Image

FelipeCybis:
- Close #179 

The **Signals plot** was resolving the x-axis dimension inconsistently for data with no `time` dimension: the slider cursor read world coordinates (mm) while the trace was falling back to step indices with a hard-coded "Time" label, so the two never aligned and the x-axis dropdown was empty.

This PR:
- `_get_available_xaxis_dims`: offer the non-displayed (slider) axes with more than one element, restoring the pre-regression criterion.
- `_get_xaxis_coords` / `_get_xaxis_label`: resolve the dimension via the shared `_xaxis_dim_index` so trace, label, and cursor agree even when the dropdown is empty.
- `_render_imported_only`: derive the label instead of hard-coding "Time".

Fixed MRE from #179:
image

## Checklist


FelipeCybis:
@sdiebolt ready to review!

sdiebolt (approved):

PR #181, opened 2026-06-15
fix(docs): make gallery build CI-friendly

### LINKED ISSUE #178: `build_gallery.py` outputs should better integrate with github CI

Today, during the build step of the gallery, we have just silent until it builds or fails. I suppose that the rich output we have locally does not translate well to the Github CI.

Also, just saying that there is a different on the number of outputs on a given cell is not super helpful. Showing the outputs would be much better.

FelipeCybis:
- Close #178

## Summary

Two fixes to the examples-gallery build tooling, prompted by an opaque CI build step.

### 1. CI-friendly progress output
The builder used a `rich` live progress bar, which emits nothing useful in non-TTY environments (which is where our actions run). The step looked **silent until it built or failed**.

Now `tools/gallery/_pipeline.py` detects whether stdout is interactive (`sys.stdout.isatty()`):
- **Interactive** → unchanged live `rich` bar.
- **Non-interactive (CI / piped)** → the bar is disabled and plain, flushed per-example lines are printed instead:
  ```
  -> io/confusius_xarray_101 (12 cells, light+dark)
     done in 4.2s
  -> qc/foo [cached]
  ```

### 2. Show outputs on a light/dark parity mismatch
When the light and dark runs produce a different *number* of outputs for a cell, the builder raised a `ValueError` that reported only the counts — unhelpful for finding the culprit. The message now lists the actual outputs on each side:
```
demo: cell 1 produced 2 light outputs and 1 dark outputs. ...
Light outputs (2):
  1. stream[stdout]: line0
  2. stream[stderr]: only-in-light warning
Dark outputs (1):
  1. stream[stdout]: line0
```
so a stray one-shot warning / download log is obvious.

## Tests
Added tests and verification of the GA output helped me fixed a non deterministic bug sometimes happening with warnings in the registration example (1885d7cc9b3ea34180815b31b4c94b81d4173d1e)

### DIFF
--- a/docs/examples/02_registration/01_register_volume_same_subject.py
+++ b/docs/examples/02_registration/01_register_volume_same_subject.py
@@ -60,10 +60,11 @@ def _load_angio_for_registration(session: str) -> xr.DataArray:
 
 
 fixed = _load_angio_for_registration(sessions[0])
-moving = _load_angio_for_registration(sessions[1])
 
 fixed
 # %%
+moving = _load_angio_for_registration(sessions[1])
+
 moving
 
 # %% [markdown]
--- a/justfile
+++ b/justfile
@@ -10,8 +10,14 @@ gallery:
     uv run python tools/build_gallery.py
 
 # Remove generated gallery artifacts and the gallery cache.
+[unix]
 clean-gallery:
-    rm -rf docs/examples/_built docs/examples/index.md .cache/gallery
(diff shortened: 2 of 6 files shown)


FelipeCybis:
@sdiebolt this is good to be reviewed! Small improvements on the example gallery building step.

sdiebolt (approved):

PR #182, opened 2026-06-17
refactor(napari): Resize ConfUSIus widget header to make more space to panels

### LINKED ISSUE #183: [Napari plugin] Signals panel does not resize like the other panels when scrollbar is present
_Originally posted by @sdiebolt in https://github.com/confusius-tools/confusius/issues/182#issuecomment-4731755945_

> I think that's a problem of the signals tab currently: See the position of the groups and Take a Tour button between the Data I/O tab and the Signals tab. Maybe we could fix that here?

> image

> image

FelipeCybis:
## Description

Small rework to make the header of the ConfUSIus widget in napari a bit smaller without loosing any of the content.
I recognize this is not massive space gained, but it is a start :P

It also fixes #183, an overflow problem in one of the plugin panels.

From:
image

To:
image


FelipeCybis:
@sdiebolt ready to review, although it will need to re-generate all docs images related to the GUI

sdiebolt:
That's indeed a good start! There might be an overflow problem though? Look at the "Take a tour" button?

image

FelipeCybis:
> That's indeed a good start! There might be an overflow problem though? Look at the "Take a tour" button?
> image

I think it is just the scroll bar. Maybe we should control for it not to appear on the screenshots. We have the same behavior today (see https://confusius.tools/latest/images/gui/plugin-signals.png)

sdiebolt:
I think that's a problem of the signals tab currently: See the position of the groups and Take a Tour button between the Data I/O tab and the Signals tab. Maybe we could fix that here?

image

image

FelipeCybis:
> I think that's a problem of the signals tab currently: See the position of the groups and Take a Tour button between the Data I/O tab and the Signals tab. Maybe we could fix that here?

Oh, indeed, I will take a look

FelipeCybis:
@sdiebolt here is the new screenshot from the Signals panel (see https://github.com/confusius-tools/confusius/issues/183#issuecomment-4741472847 for what the problem was)

image

sdiebolt (approved):
LGTM!

PR #187, opened 2026-06-19
fix(xarray): anchor `physical_to_qform` to the primary (sform) physical frame

### LINKED ISSUE #184: `physical_to_qform` is detached from the physical frame
In [`confusius/io/nifti.py::_create_spatial_coords_from_nifti`](https://github.com/confusius-tools/confusius/blob/15b0648213cbacbb37bdb357e2b22c9076a72c72/src/confusius/io/nifti.py#L267), the NIfTI sform and qform are processed independently:

- The **primary** affine (sform when `sform_code > 0`, else qform) defines the physical coordinates: `coord = T_primary + Z_primary · index`, where `T, Z` come from `decompose_affine`. The orientation `D = A[:3,:3] / Z` (rotation + shear) is stripped from the 1D coords and stored in `physical_to_{primary}` so that `physical_to_{primary} @ physical_pos = primary_world_pos`.

- `physical_to_qform` (the secondary) is built from **qform's own** decomposition: `offset = T_qform − D_qform · T_qform`. This implicitly assumes `physical_pos = T_qform + Z_qform · index`, which is false, because the real physical coordinates are the sform's. Consequently `physical_to_qform @ physical_pos ≠ qform_world_pos`: the stored qform affine has lost its relationship to the actual physical coordinate frame.

**-> A nasty side-effect of this is that a round trip with a NIfTI file containing `sform` and `qform` actually silently breaks the `qform`!**

## Expected

In a NIfTI file with `qform` and `sform`, both `physical_to_[q|s]form` should match the physical coordinates of the DataArray.

## Actual

A file with valid `[q|s]form` have `physical_to_[q|s}form` affine that does not match the physical coordinate space.

FelipeCybis:
- Fixes #184

## Summary

ConfUSIus DataArrays carry an `attrs["affines"]` dict whose 4x4 matrices may map the array's physical coordinates (its 1D `z`/`y`/`x` dimension coordinates) into various 'world' frames.

When `load_nifti` read a file with both `sform` and `qform`, the physical coordinates were built from the primary (`sform`) affine, but `physical_to_qform` was computed from `qform`'s own decomposition, implicitly assuming the coordinates came from `qform`. They do not, so `physical_to_qform @ physical_pos` did not equal `qform_world`.

This PR fixes the `physical_to_qform` to correctly link to the physical coordinates when they are built with the `sform`.
The physical frame is now defined once by the primary affine, and every form (primary and secondary) is re-expressed against that shared frame. `physical_to_sform` and the coordinates are unchanged, and `physical_to_qform` is now correctly anchored.

It also introduces the shared `axis_aligned_affine` / `reexpress_affine` helpers used to re-express an affine relative to a physical frame. (The follow-up PR #188 builds on these helpers to let `apply_affine` accept rotation/shear.)

## Addendum

This PR revealed a more nasty bug: `qform`s were silently corrupted in a NIfTI file roundtrip with ConfUSIus. This PR fixes this by refactoring the NIfTI saving functions so it reverses the exact loading logic.

## Checklist


FelipeCybis:
@sdiebolt this is ready to be reviewed!
You will see that working on this I found that a `qform` gets corrupted in a round-trip if the NIfTI file has a `sform` as well. This is quite nasty, IMHO, so we should try to get a new release soon after this.

docs/user-guide/spatial-conventions.md
```diff
@@ -204,5 +208,7 @@ fixed/reference DataArray are preserved automatically.
     preserves coordinates, so the affine remains valid without any adjustment.
 
     [`save_nifti`][confusius.io.save_nifti] reconstructs the full voxel → reference
-    NIfTI affine internally by combining the physical → reference orientation matrix
-    with the dimension coordinate origin and spacing.
+    NIfTI affine internally by composing the physical → reference affine with the
+    voxel → physical map built from the dimension coordinate origin and spacing. Using
+    the full affine (not just its orientation) means each form keeps its own origin, so
```
sdiebolt:
```suggestion
    the full affine (not just its orientation) means each affine keeps its own origin, so
```

src/confusius/_utils/coordinates.py
```diff
@@ -274,3 +274,63 @@ def get_coordinate_origins(data: xr.DataArray) -> dict[str, float]:
...
+    (4, 4) numpy.ndarray
+        Homogeneous affine with `diag(zoom)` and `translation` as its last column.
+    """
+    affine = np.eye(4)
+    affine[:3, :3] = np.diag(np.asarray(zoom, dtype=np.float64))
+    affine[:3, 3] = np.asarray(translation, dtype=np.float64)
+    return affine
+
+
+def reexpress_affine(
```
sdiebolt: I think we should try to find a clearer name for this function. Maybe `reexpress_affine_to_axis_aligned`?

src/confusius/io/nifti.py
```diff
@@ -274,8 +276,11 @@ def _create_spatial_coords_from_nifti(
     Selects the primary affine (sform preferred over qform when both are valid)
     and builds one coordinate per spatial dimension. When a valid affine is
     available, each axis starts at `affine[col, 3]` (origin) and is stepped by
-    the true voxel size (from `decompose_affine`). The rotation and shear are
-    captured in a 4×4 `A_physical` affine stored in the returned attributes.
+    the true voxel size (from `decompose_affine`). The physical coordinate frame
+    is defined by the primary affine's translation and zoom; every form's affine
```
sdiebolt:
```suggestion
    is defined by the primary affine's translation and zoom; every affine
```

src/confusius/io/nifti.py
```diff
@@ -333,6 +338,14 @@ def _create_spatial_coords_from_nifti(
         _, sform_code = img.header.get_sform(coded=True)
         primary_prefix = "sform" if sform_code > 0 else "qform"
 
+        # The physical coordinate frame is defined once by the primary affine's
+        # axis-aligned decomposition: origin T (translation) and voxel spacing Z (true
+        # zoom from decompose_affine). Every stored affine is re-expressed against this
+        # single frame, so a secondary form (e.g. qform when sform is primary) keeps its
```
sdiebolt:
```suggestion
        # single frame, so a secondary affine (e.g. qform when sform is primary) keeps its
```

sdiebolt (approved):
Nice fix! Added a few comments about semantics. We should indeed do a quick fix release after this is merged.

src/confusius/_utils/coordinates.py
sdiebolt: Tried to clarify in [3515dff](https://github.com/confusius-tools/confusius/pull/187/commits/3515dffb4c5904a171d0b83e9290b695fd6c0bd7)

PR #188, opened 2026-06-21
feat(xarray): `apply_affine` accepts rotation/shear and returns the residual orientation

### LINKED ISSUE #185: `apply_affine` should accept affines with rotation/shear and create a new attribute affine to store the rotation/shear
[`confusius/xarray/affine.py::apply_affine`](https://github.com/confusius-tools/confusius/blob/15b0648213cbacbb37bdb357e2b22c9076a72c72/src/confusius/xarray/affine.py#L65) raises `ValueError` whenever the provided affine's rotation block is non-diagonal. Meaning that it only handles axis-aligned transforms (scale + translation). 

It should instead mirror the [NIfTI loader](https://github.com/confusius-tools/confusius/blob/15b0648213cbacbb37bdb357e2b22c9076a72c72/src/confusius/io/nifti.py#L267): decompose the affine, absorb the axis-aligned part into the 1D coordinates, and store the residual orientation as a new physical-frame affine so the full transform is preserved. The rest of the stored affines are already correctly updated to the new physical space.

FelipeCybis:
- Fixes #185

> [!NOTE]
> ~Stacked on #187 (this PR's base is `worktree-physical-affine-frame`). It builds on the shared `reexpress_affine` helper introduced there. GitHub will retarget this PR to `main` automatically once #187 merges.~

## Summary

This PR extends the translation/zoom/rotation/shear decomposition scheme (introduced for the NIfTI loader in #187) to `apply_affine`, allowing it to accept transforms that contain rotations and shears.

`data.fusi.affine.apply` previously raised on any non-diagonal affine. It now returns a `(DataArray, orientation)` tuple:

- the axis-aligned part (per-axis zoom and translation) is absorbed into the 1D `z`/`y`/`x` coordinates, and
- the residual orientation that the coordinates cannot represent is returned as a 4x4 affine mapping the new physical coordinates to the transform's target world frame (`orientation @ new_physical == affine @ old_physical`). It is the identity for a diagonal affine.

Returning the orientation, rather than forcing a caller-provided `key`, lets the caller decide what to do with it (compose it with a stored affine, store it under a key of their choosing, or ignore it). This matters when the goal is only to re-express the coordinates (for example into the `qform` frame) and the relevant affine already exists in `attrs["affines"]`. All existing stored affines are re-expressed so they stay valid in the new frame.

## Checklist


FelipeCybis:
@sdiebolt This is ready to review as well. I added a section in the user guide under "Spatial convention" that might be worth a more thorough review.

docs/user-guide/spatial-conventions.md
```diff
@@ -210,5 +210,147 @@ fixed/reference DataArray are preserved automatically.
     [`save_nifti`][confusius.io.save_nifti] reconstructs the full voxel → reference
     NIfTI affine internally by composing the physical → reference affine with the
     voxel → physical map built from the dimension coordinate origin and spacing. Using
-    the full affine (not just its orientation) means each affine keeps its own origin, so
+    the full affine (not just its orientation) means each form keeps its own origin, so
```
sdiebolt:
```suggestion
    the full affine (not just its orientation) means each affine keeps its own origin, so
```

sdiebolt (approved):
LGTM!

PR #190, opened 2026-06-22
perf(iq): use `map_blocks` for non-overlapping IQ windows

### LINKED ISSUE #189: Optimize non-overlapping IQ block processing with `map_blocks`
## Summary

`confusius.iq.process_iq_blocks()` currently uses `dask.array.map_overlap()` for all cases. When the outer IQ windows do not overlap (`window_stride == window_width`), that adds avoidable overlap-task overhead.

## Proposed change

Use `dask.array.map_blocks()` as a dedicated fast path when `window_width - window_stride == 0`, while keeping the existing `map_overlap()` path for overlapping windows.

## Motivation

For non-overlapping outer windows, each output block depends on exactly one input block, so the overlap machinery is unnecessary.

## Initial benchmark signal

On a real IQ Zarr dataset, this reduced task count in the non-overlapping case and produced a modest but noticeable end-to-end improvement:

- light mean reducer: ~0.0318 s -> ~0.0269 s
- `compute_power_doppler_volume`: ~14.07 s -> ~11.61 s

The exact gain depends on scheduler/kernel details, but the fast path appears worthwhile and low risk.

sdiebolt:
## Summary

Use `dask.array.map_blocks()` in `process_iq_blocks()` when the outer IQ windows do not overlap (`window_stride == window_width`). Keep the existing `map_overlap()` path for overlapping windows.

## Changes

- add a dedicated `map_blocks()` fast path for `overlap_width == 0`
- keep the existing truncation/warning logic unchanged
- add tests covering:
  - non-overlapping windows using the fast path
  - `drop_axis` behavior on the fast path
  - overlapping windows still using `map_overlap()`

Closes #189


FelipeCybis (approved):
LGTM after a changelog entry ;)

PR #192, opened 2026-06-23
fix(iq): handle strongly overlapping IQ windows

### LINKED ISSUE #191: `process_iq_blocks` mis-handles overlapping windows when `window_stride < window_width / 2`
When `process_iq_to_power_doppler()` is called with strongly overlapping clutter windows, xarray can fail with a `CoordinateValidationError` because the processed data length and computed `time` coordinate length diverge.

Example:
```python
iq_chunk.fusi.iq.process_to_power_doppler(
    clutter_window_width=200,
    clutter_window_stride=1,
    doppler_window_width=10,
    low_cutoff=30,
).compute()
```

Error:
```python
CoordinateValidationError: conflicting sizes for dimension 'time':\nlength 500 on the data but length 96020 on coordinate 'time'\n```\n\nThis happens in the overlapping path of `process_iq_blocks()`: time is rechunked as `(window_width, window_stride, window_stride, ...)`, then `dask.array.map_overlap()` is called with left overlap `window_width - window_stride`.\n\nThis only works as intended when `window_stride >= window_width / 2`. For smaller strides, Dask auto-rechunks to satisfy the requested overlap. That breaks the assumption that each input block produces the same output shape as the first one, so Dask records an incorrect output time dimension while `compute_processed_volume_timings()` still computes the full number of windows.\n\nExpected behavior: smaller strides should be supported correctly, limited only by compute/memory costs, not by chunking artifacts.

sdiebolt:
## Summary
- batch overlapping IQ windows before applying Dask overlap
- keep output chunk sizes aligned with the true number of sliding-window outputs
- add regression tests for overlapping `process_iq_blocks` and chunked power Doppler processing

Closes #191.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -19,6 +19,10 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- `process_iq_blocks` now handles strongly overlapping IQ windows without corrupting
+  the output time dimension, so power Doppler and related IQ reducers work when
+  `window_stride < window_width / 2`
+  ([#192](https://github.com/confusius-tools/confusius/pull/192)).
 - `load_nifti` now anchors `physical_to_qform` to the same physical frame as the
   primary (sform) coordinates, so the stored qform affine maps the array's
   physical coordinates to qform world space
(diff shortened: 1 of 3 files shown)


FelipeCybis (approved):
I think I would need to be working more closely to beamformed IQ data to be of more help here. But this looks good to me!

PR #197, opened 2026-06-24
fix: coerce mask to boolean after validation

### LINKED ISSUE #196: Masks should always be coerced to `bool` before using as indexer
### ConfUSIus version

v0.3.1.dev

### Python version

3.14

### Operating system

WSL

### Installation

using uv

### Description

When `DataArray` is indexed (`isel`) with an integer array. A [positional indexing](https://docs.xarray.dev/en/stable/user-guide/indexing.html#positional-indexing) is used. Meaning that an array `A` with 6 elements indexed with `[0, 0, 1, 1, 0, 0]` will return a 6-element array where the first two are the 0th element of `A`, the third and fourth element will be the 1st element of `A` and the final two elements will be again the 0th element of `A`.

This is controlled for in some places:

https://github.com/confusius-tools/confusius/blob/473114f5788c82a32a2ffbe5cae9feb93c57785f/src/confusius/extract/mask.py#L108-L110

https://github.com/confusius-tools/confusius/blob/473114f5788c82a32a2ffbe5cae9feb93c57785f/src/confusius/iq/clutter_filters.py#L772-L774

But not others:

https://github.com/confusius-tools/confusius/blob/473114f5788c82a32a2ffbe5cae9feb93c57785f/src/confusius/iq/process.py#L1189-L1191

https://github.com/confusius-tools/confusius/blob/473114f5788c82a32a2ffbe5cae9feb93c57785f/src/confusius/iq/process.py#L1538-L1540

https://github.com/confusius-tools/confusius/blob/473114f5788c82a32a2ffbe5cae9feb93c57785f/src/confusius/signal/confounds.py#L542-L552

This last one is particularly funny because `selected_voxels = selected_voxels & noise_mask_flat` is a bitwise operation and if the non-zero value of the `noise_mask_flat` is even, `selected_voxels` becomes all zero, if it is odd, `selected_voxels` comes back as 1 where `True & odd`. and zero elsewhere. 

### Steps to reproduce

```python
"""Minimal reproducible example: `compute_compcor_confounds` ignores integer label masks.

`Atlas.get_masks(region)` returns an INTEGER mask whose foreground voxels carry the
region id (e.g. 1009) and whose background is 0 -- it is NOT a boolean mask. Feeding
such a mask to `compute_compcor_confounds` used to select the wrong voxels, because
`bool & int` is a bitwise AND that yields an integer array, which `xarray.isel()` then
treats as *positional* indices instead of a boolean selector.

To make the failure observable through the public API, we plant a clean oscillation
ONLY in the masked region. A correct aCompCor restricted to that region must recover it
as its first component; an incorrect one (that pulled in the wrong voxels) cannot.
"""
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#
# for issues.

import numpy as np
import xarray as xr

from confusius.signal import compute_compcor_confounds

rng = np.random.default_rng(0)
n_time, n_voxels = 200, 40

# Independent noise everywhere, plus a planted oscillation in the noise region only.
t = np.arange(n_time)
region_signal = np.sin(2 * np.pi * t / 25.0)
data = rng.standard_normal((n_time, n_voxels))
region = slice(30, n_voxels)  # the 10 "white-matter" voxels we want to mask
data[:, region] += 5.0 * region_signal[:, None]
fusi = xr.DataArray(data, dims=["time", "space"], coords={"time": t.astype(float)})

# Integer label mask, exactly as `Atlas.get_masks(1009)` produces it: values in {0, 1009}.
int_mask = xr.DataArray(np.zeros(n_voxels, dtype=np.int32), dims=["space"])
int_mask.values[region] = 1009

# The boolean mask it is *supposed* to behave identically to.
bool_mask = int_mask.astype(bool)

comp_int = compute_compcor_confounds(fusi, int_mask, n_components=1)
comp_bool = compute_compcor_confounds(fusi, bool_mask, n_components=1)


def recovery(component: xr.DataArray) -> float:
    """Absolute correlation of the first component with the planted region signal."""
    return abs(np.corrcoef(component.values[:, 0], region_signal)[0, 1])


print(f"integer label mask : |corr with planted signal| = {recovery(comp_int):.3f}")
print(f"boolean mask       : |corr with planted signal| = {recovery(comp_bool):.3f}")
print(
    f"integer == boolean : {np.allclose(np.abs(comp_int.values), np.abs(comp_bool.values))}"
)

```

Output:
```
integer label mask : |corr with planted signal| = 0.038
boolean mask       : |corr with planted signal| = 0.995
integer == boolean : False
```

### Additional output

_No response_

FelipeCybis:
- Closes #196 

## Description

Indexing DataArrays with masks the way we do regularly in ConfUSIus demands a boolean mask. Instead of transforming the `dtype` of the mask every time after validation, `validate_mask` now return the input mask. The return type can be controlled with `return_dtype_as_bool` which is `True` as a default.

This fixes several points where the mask could be used as an indexer with an `int` dtype, 

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,9 +27,9 @@ Current development version for the next ConfUSIus release.
 ### :bug: Fixes
 
 - Masks are now coerced to boolean by `validate_mask` (added `return_dtype_as_bool`
-  parameter that defaults to `True`), to avoid DataArrays to use *positional indexing*
-  instead. Previously these masks could select the wrong voxels or, for
-  `register_volume`, silently disable the metric mask
+  parameter that defaults to `True`) to avoid DataArrays using *positional indexing*.
+  Previously these masks could select the wrong voxels or, for `register_volume`,
+  silently disable the metric mask
   ([#197](https://github.com/confusius-tools/confusius/pull/197)).
 - `load_nifti` now anchors `physical_to_qform` to the same physical frame as the
   primary (sform) coordinates, so the stored qform affine maps the array's
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -19,6 +19,11 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- Masks are now coerced to boolean by `validate_mask` (added `return_dtype_as_bool`
+  parameter that defaults to `True`), to avoid DataArrays to use *positional indexing*
+  instead. Previously these masks could select the wrong voxels or, for
+  `register_volume`, silently disable the metric mask
+  ([#197](https://github.com/confusius-tools/confusius/pull/197)).
 - `load_nifti` now anchors `physical_to_qform` to the same physical frame as the
   primary (sform) coordinates, so the stored qform affine maps the array's
   physical coordinates to qform world space
--- a/src/confusius/decomposition/_base.py
+++ b/src/confusius/decomposition/_base.py
@@ -264,15 +264,14 @@ class _BaseFUSIDecomposer(BaseEstimator, TransformerMixin):
 
         X_ordered = X.transpose("time", *spatial_dims)
 
-        if getattr(self, "mask", None) is not None:
-            mask = self.mask
-            assert mask is not None
-            validate_mask(mask, X_ordered, "mask", require_exact_dims=True)
-        else:
+        mask = getattr(self, "mask", None)
+        if mask is None:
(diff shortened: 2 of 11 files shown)


FelipeCybis:
@sdiebolt ready to review!

sdiebolt (approved):

PR #204, opened 2026-06-24
feat: add `apply_statistical_threshold` with multiple comparison correction methods

### LINKED ISSUE #200: Add a "thresholding" function with multiple comparison correction methods.

We should have something like https://nilearn.github.io/dev/modules/generated/nilearn.glm.threshold_stats_img.html
so we can easily find statistical thresholds after analysis.

FelipeCybis:
- Close #200 

## Description

Add `apply_statistical_threshold` for either:

- Threshold a DataArray with a provided `threshold`.
- Find the statistical threshold given an `alpha` and one of the mulitple comparison methods:
  - `fpr` for uncorrected false positive rate
  - `fdr_bh` for the false discovery rate (Benjamini-Hochberg)
  - `bonferroni` for the Bonferroni correction

## Checklist


FelipeCybis:
@sdiebolt ready to be reviewed!

FelipeCybis:
@sdiebolt now this is ready to be reviewed :)

FelipeCybis:
@sdiebolt the other option would be to incorporate `statsmodels` and use their [multiple tests](https://www.statsmodels.org/dev/_modules/statsmodels/stats/multitest.html#multipletests) methods (a few more than what is implemented here).

sdiebolt:
Possibly we could also just reimplement the necessary methods and adding their license to avoid having a new dependency! But that could go in a later PR.

sdiebolt:
@FelipeCybis Wondering if `glm` is the right subpackage for this... It could also be applied to any kind of statistical maps, e.g. seed-based maps, right? So shouldn't we create a `stats` subpackage?

FelipeCybis:
> @FelipeCybis Wondering if `glm` is the right subpackage for this... It could also be applied to any kind of statistical maps, e.g. seed-based maps, right? So shouldn't we create a `stats` subpackage? 

@sdiebolt indeed. I have no objection to that.

sdiebolt:
Ok done. Now second thing I'm wondering about: `apply_statistical_threshold` acts specifically on z-maps. But there are a few other statistics that could benefit from such a function. Notably, most multiple comparison correction function are typically applied directly on p-values rather than on z-scores... So I'm wondering if we should maybe modify this function to act on a DataArray of p-values, and maybe return e.g. an alpha map or a threshold rather than thresholding the DataArray directly?  

**edit:** actually maybe we could split the function and extract the multiple comparison part into `adjust_pvalues`?

FelipeCybis:
> Ok done. Now second thing I'm wondering about: `apply_statistical_threshold` acts specifically on z-maps. But there are a few other statistics that could benefit from such a function. Notably, most multiple comparison correction function are typically applied directly on p-values rather than on z-scores... So I'm wondering if we should maybe modify this function to act on a DataArray of p-values, and maybe return e.g. an alpha map or a threshold rather than thresholding the DataArray directly?

We could offer a `zscore` mode and `pvalue` mode if that is what you are implying it would be useful. In the `pvalue` mode the returned DataArray would be the corrected p-values and the threshold would not actually be needed but to keep two outputs in both modes we could return the untouched alpha provided by the user. Is this in line with what you are thinking?

I did wonder if splitting the function should be done. Maybe we could have part of the private functions made public so they can be used separately.

sdiebolt:
@FelipeCybis what do you think of the split I've done?

FelipeCybis:
> @FelipeCybis what do you think of the split I've done?

Looks good to me! Good to merge!

sdiebolt (approved):

PR #206, opened 2026-06-25
feat(cli): support opening multiple files in a single invocation

### LINKED ISSUE #205: CLI: support opening multiple files for batch operations
The ConfUSIus CLI currently only accepts a single file at a time, which makes batch workflows (e.g. running registration across many acquisitions) awkward. Allowing multiple files to be passed in a single invocation would unlock common pipelines directly from the terminal.

sdiebolt:
## Summary

- Allow `confusius` CLI to accept multiple positional file paths so users can open batches (e.g. fixed/moving pairs for registration) in one call.
- Each path becomes its own napari image layer, named after the file's basename.
- Split the parser out into a public `build_parser()` and the side-effecting work into a public `run()` so the CLI is testable against a headless napari viewer.

## Usage

```bash
confusius fixed.nii moving.nii
confusius data/*.nii --lazy
confusius scan.scan --video cam.mp4
```

## Test plan

- [x] `uv run pytest tests/unit/test_cli.py` — 9 new tests, all green
- [x] `just pre-commit` — ruff check/format, ty, codespell, numpydoc all pass
- [x] CI

## Related

Closes #205

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,10 @@ Current development version for the next ConfUSIus release.
 
 ### :sparkles: Enhancements
 
+- The `confusius` CLI now accepts multiple fUSI data files in a single
+  invocation (e.g. `confusius fixed.nii moving.nii`). Each file is added as its
+  own image layer, named after the file's basename
+  ([#205](https://github.com/confusius-tools/confusius/issues/205)).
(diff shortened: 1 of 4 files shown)


FelipeCybis (approved):
Added a hint in the GUI overview part of the docs. LGTM!

PR #222, opened 2026-06-29
fix(nifti): preserve sidecar affines when loading

### LINKED ISSUE #221: load_nifti drops sidecar `affines` entries (e.g. `bspline_initialization`)
## Bug

`confusius.io.load_nifti` (and therefore `confusius.load`) silently drops affines stored in the BIDS-style JSON sidecar. Sidecar keys under `ConfUSIusAffines` (e.g. `bspline_initialization` for registration outputs) are not present on the returned `DataArray.attrs[affines]`, only the qform/sform affines from the NIfTI header.

## Repro

With a sidecar like:

```json
{
  "type": "bspline_transform",
  "order": 3,
  "direction": [[1, 0, 0], [0, 1, 0], [0, 0, 1]],
  "ConfUSIusAffines": {
    "bspline_initialization": [[1.0, 0, 0, 0], [0, 0.998, -0.058, -0.058], [0, 0.058, 0.998, 0.722], [0, 0, 0, 1]]
  }
}
```

```python
import confusius as cf
da = cf.load("bspline.nii")
assert "bspline_initialization" in da.attrs["affines"]  # FAILS
```

## Cause

In `src/confusius/io/nifti.py`, `_create_spatial_coords_from_nifti` builds `extra_attrs["affines"]` from the NIfTI header (qform/sform only). `load_nifti` then does `attrs.update(affine_attrs)`, which **replaces** the sidecar-loaded `affines` dict instead of merging into it.

sdiebolt:
Fixes #221.

`load_nifti` was overwriting the sidecar-loaded `affines` dict (e.g. `bspline_initialization` from registration outputs) with the qform/sform affines derived from the NIfTI header. The two are now merged, with NIfTI-header keys taking precedence.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -21,6 +21,13 @@ Current development version for the next ConfUSIus release.
   `data.fusi.register` accessor
   ([#215](https://github.com/confusius-tools/confusius/pull/215)).
 
+### :bug: Fixes
+
+- `load_nifti` no longer drops affines loaded from the JSON sidecar (e.g.
+  `bspline_initialization` written by the registration pipeline) when merging in the
+  NIfTI qform/sform affines
+  ([#222](https://github.com/confusius-tools/confusius/pull/222)).
+
 ## 0.4.0
 
 *Released 2026-06-25.*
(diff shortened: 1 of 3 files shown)


FelipeCybis (approved):
LGTM

PR #223, opened 2026-06-29
fix(nifti): always reserve NIfTI 4th axis for time

### LINKED ISSUE #219: `save_nifti` saves non-time axis as time
`save_nifti` currently maps the non-time additional axes to time unconditionally. This breaks non-image payloads that legitimately use additional axes for something else (for example B-spline transform control grids where the leading/component axis gets serialized through NIfTI and then comes back as time).

sdiebolt:
Fixes #219.

## Problem

A non-time 4D payload (e.g. a B-spline control grid with dims `(component, z, y, x)`) was previously serialized as a 4D NIfTI with the non-time `component` axis on NIfTI axis 4, breaking the NIfTI convention that axis 4 is the time slot. On reload, that axis was then mislabeled as `time`.

## Fix

`save_nifti` now always inserts a degenerate length-1 time axis at the 4th NIfTI slot, so non-time additional axes always land at NIfTI axes 5, 6, 7. The original dim name for each additional axis is always written to the sidecar as `ConfUSIusDim{N}Name` (with `N` in 4, 5, 6, matching the 0-based NIfTI axis of the extra dim). The matching `ConfUSIusDim{N}Coordinates` entry is only written when the coord cannot be reconstructed from `pixdim` (i.e. when the coord does not start at 0 with regular spacing); otherwise the spacing is stored in `pixdim` and the coord is rebuilt as `step * arange(size)` on load. `load_nifti` falls back to the `pixdim` reconstruction when no `ConfUSIusDim{N}Coordinates` is present, and to plain `arange(size)` when `pixdim` is zero. When the sidecar is missing the `ConfUSIusDim{N}Name` entry, the axis defaults to a `dim{N}` placeholder.

`save_nifti` now raises `ValueError` when the DataArray has more than 3 non-spatial, non-time dims (the NIfTI 7-axis limit).


FelipeCybis (approved):
LGTM

PR #225, opened 2026-06-29
fix(plotting): default alpha to None so cmap alpha is respected

### LINKED ISSUE #224: Having `alpha` as 1 in matplotlib plotting functions blocks the alpha channel of a colormap
We have `alpha` defaulting to 1 in the matplotlib-based plotting. It should be `None`, otherwise the `alpha` channel of a cmap does not work.

FelipeCybis:
## Summary

The matplotlib image plotters defaulted `alpha` to `1.0`, which forces full opacity and overrides a colormap's own alpha channel. This defaults `alpha` to `None` (the matplotlib-native value) so a colormap's built-in alpha is respected.

Affected: `plot_volume`, `plot_composite`, the `VolumePlotter.add_volume` / `add_composite` methods, and the `.fusi.plot_volume` / `.fusi.plot_composite` accessors.

## Test

Added `test_default_alpha_preserves_cmap_alpha`: renders a volume with a semi-transparent `ListedColormap` and asserts the QuadMesh facecolors keep the colormap's alpha (the old `alpha=1.0` default flattened them to `1.0`).

Closes #224

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -23,6 +23,9 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- Image plotting functions now leave `alpha` unset by default (`None`), so a
+  colormap's built-in alpha channel is respected
+  ([#225](https://github.com/confusius-tools/confusius/pull/225)).
 - `load_nifti` no longer drops affines loaded from the JSON sidecar (e.g.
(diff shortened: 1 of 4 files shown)


FelipeCybis:
@sdiebolt ready to be reviewed!

sdiebolt (approved):

PR #235, opened 2026-07-02
feat(registration): add displacement-field sampling and inversion for B-spline transforms

### LINKED ISSUE #229: Support B-spline transforms in Atlas.resample_like
`confusius.registration.resample_like` accepts both affine (`numpy.ndarray`) and B-spline (`xarray.DataArray`) transforms, but the wrapper `Atlas.resample_like` only accepts an affine matrix (`src/confusius/atlas/atlas.py:513-520`, typed `npt.NDArray[np.float64]`).

This blocks users who registered an atlas with `register_volume(transform="bspline")` from resampling the atlas with the resulting transform. The docstring explicitly states that `Atlas.resample_like` is meant to mirror `confusius.registration.resample_like`, so the asymmetry is unintentional.

Meshes returned by `get_mesh` should follow the same deformation as the volume to stay useful as overlays — meaning the affine assumption baked into the mesh-to-physical update (`src/confusius/atlas/atlas.py:595`) needs to be lifted in the process.

### LINKED ISSUE #233: Add helper to invert B-spline transforms via displacement-field inversion
`confusius.registration` exposes `resample_like`, which applies a transform in the forward direction, but no symmetric helper to apply its inverse. Affine transforms can be inverted analytically (`np.linalg.inv`), but B-spline transforms cannot — they have no closed-form inverse, which blocks use cases like "load a saved registration transform, apply its inverse to a layer that lives in the fixed/reference space" (currently infeasible from the napari plugin because `apply_selected_transform` is unidirectional and the transform payload stores no input grid either).

The standard recipe is well supported by SimpleITK:

1. Sample the B-spline into a dense displacement field on a chosen grid via `sitk.TransformToDisplacementFieldFilter` (set size/spacing/origin from the moving or fixed image).
2. Iteratively invert the field via `sitk.InvertDisplacementFieldImageFilter` (fixed-point composition, ~10–20 iterations, `MaxErrorToleranceThreshold≈0.1`).
3. Either wrap the inverse field in a `sitk.DisplacementFieldTransform` for direct use, or fit a new B-spline to it via `sitk.BSplineTransform` if a compact control-point representation is needed.

This would fit naturally alongside `sitk_bspline_to_dataarray` in `src/confusius/registration/bspline.py` (e.g. `invert_bspline_dataarray`), and a corresponding napari hook could be layered on top later.

Caveats to document: works best for small/diffeomorphic deformations, requires positive Jacobian in the region of interest, and `EnforceBoundaryCondition` clamps the inverse to zero at the field boundary by default — relevant if the field extends past the target grid.

Refs: SimpleITK `TransformToDisplacementFieldFilter`, `InvertDisplacementFieldImageFilter`, `DisplacementFieldTransform`.

### LINKED ISSUE #236: B-spline control-point grid geometry is swapped between axes on anisotropic images
`sitk_bspline_to_dataarray` and its inverse `_dataarray_to_sitk_bspline` in `src/confusius/registration/bspline.py` assumed SimpleITK reverses axis order relative to the DataArray (`(x, y, z)` vs. `(z, y, x)`) and applied a permutation to compensate. In fact ConfUSIus' actual convention (established by `dataarray_to_sitk_image`, and used consistently everywhere else, e.g. `resample_volume`) never reverses axes: SimpleITK axis `i` maps directly to DataArray dim `i`.

sdiebolt:
## Summary

- Adds `bspline_to_displacement_field` and `invert_displacement_field` to `confusius.registration.bspline`: sample a B-spline (or composite affine + B-spline) transform into a dense displacement field on an explicit grid, then invert it via SimpleITK's `InvertDisplacementFieldImageFilter`.
- `resample_volume`/`resample_like` now accept displacement fields directly alongside affines and B-spline DataArrays, so a saved registration transform's inverse can be applied without a closed-form inverse — the missing piece for a future napari "apply inverse transform" hook.
- Fixes two bugs found while validating against real (anisotropic, singleton-axis) fUSI data:
  - `sitk_bspline_to_dataarray`/`_dataarray_to_sitk_bspline` assumed SimpleITK reverses axis order relative to the DataArray; the rest of the module never reverses, so this silently swapped control-point grid geometry between axes on anisotropic images (invisible on isotropic test fixtures). Pre-existing, shipped in released alpha versions — see #236.
  - `InvertDisplacementFieldImageFilter` silently returns an all-zero field when any spatial axis has size 1 (fUSI data is routinely a single 2D slice, e.g. `(1, y, x)`). `invert_displacement_field` now reuses `expand_thin_dims` (promoted from `volume.py` to the shared `registration/_utils.py`) to pad and crop degenerate axes — the same trick `register_volume` already relies on for its own thin-dimension images. New code introduced in this PR, not a prior regression.
- Extends the same-subject registration example with a B-spline refinement step after the rigid one.
- Adds atlas mesh warping/resampling support for affine and nonlinear transforms, plus unit-validation coverage for atlas and registration resampling paths.

Closes #229. Closes #233. Closes #236.


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (commented):
I am a bit confused about the `_mesh_vertex_transform`. It says on the `Atlas` docstring that it is a pull transform that maps the atlas' current physical coordinates back to the base atlas physical coordinates. The `from_brainglobe` constructor sets it as the identity so this makes sense with the description as BrainGlobe atlases are oriented in
`asr` and we accept this as `zyx` and mesh coordinates are also in this orientation. The affine transformations when `_apply_mesh_vertex_transform` are fine, but if a displacement field is given, in `_interpolate_displacement_field` we have the grid from the displacement field coordinates `("component", "z", "y", "x")` and the points that are given in the base atlas coordinates. Except here the docstring says `Physical points in atlas mesh order (x, y, z) (or (x, y) in 2D).` (shouldn't this be base atlas physical space?). The code inverts the second dimension of the points, interpolate with the displacement field and then inverts the second dimension of the result back. Shouldn't the displacement field components also
correspond to `0=z`, `1=y` and `2=x`? This reverberates through the tests, for example:

```python
    def test_accepts_displacement_field_and_warps_mesh(self, atlas: Atlas) -> None:
        reference = xr.DataArray(
            np.zeros((4, 3, 4), dtype=np.float32),
            dims=["z", "y", "x"],
            coords={
                "z": xr.Variable("z", [0.0, 0.05, 0.1, 0.15], attrs={"voxdim": 0.05}),
                "y": xr.Variable("y", [0.0, 0.05, 0.1], attrs={"voxdim": 0.05}),
                "x": xr.Variable("x", [0.0, 0.05, 0.1, 0.15], attrs={"voxdim": 0.05}),
            },
        )
        dims = list(reference.dims)
        coords = {dim: reference.coords[dim] for dim in dims}

        field = xr.DataArray(
            np.zeros((3, *reference.shape), dtype=np.float64),
            dims=["component", *dims],
            coords={"component": np.arange(3), **coords},
            attrs={"type": "displacement_field_transform"},
        )
        field.loc[dict(component=2)] = -0.01

        resampled = atlas.resample_like(reference, field)
        vertices_mm, _ = resampled.get_mesh(997)

        expected = np.array(
            [
                [0.01, 0.0, 0.05],
                [0.11, 0.0, 0.05],
                [0.01, 0.1, 0.05],
                [0.01, 0.0, 0.15],
                [0.11, 0.0, 0.15],
                [0.01, 0.1, 0.15],
            ]
        )
        np.testing.assert_allclose(vertices_mm, expected, atol=1e-6)
```

where the displacement is in component 2 and the expected result is in the first column of the vertices. I think this is coming from `_interpolate_displacement_field`:

```python
    spatial_dims = [str(dim) for dim in field.dims[1:]]
    grid = tuple(
        np.asarray(field.coords[dim].values, dtype=np.float64) for dim in spatial_dims
    )
    points_for_field = points[:, ::-1]
    displacements = interpn(
        grid,
        np.moveaxis(np.asarray(field.values, dtype=np.float64), 0, -1),
        points_for_field,
        bounds_error=False,
        fill_value=np.nan,
    )[:, ::-1]
```
if points are in base atlas physical space, why invert the points?

FelipeCybis:
Ok, bug is confirmed:

```python
import numpy as np
import xarray as xr

from confusius.atlas import Atlas

atlas = Atlas.from_brainglobe("allen_mouse_100um")
reference = atlas.reference
dims = [str(d) for d in reference.dims]  # ("z", "y", "x")
print("atlas dims:", dims, "shape:", tuple(reference.sizes[d] for d in dims))


values = np.zeros((len(dims), *reference.shape), dtype=np.float64)
values[dims.index("x")] = 1
field = xr.DataArray(
    values,
    dims=["component", *dims],
    coords={
        "component": np.arange(len(dims)),
        **{d: reference.coords[d] for d in dims},
    },
    attrs={"type": "displacement_field_transform"},
)

resampled = atlas.resample_like(reference, field)

# No resampling, mesh vertices and volume are aligned
verts_before, faces_before = atlas.get_mesh("SSp")
roi_mask_before = atlas.get_masks("SSp")
viewer, _ = roi_mask_before.fusi.plot()
layer = viewer.add_surface((verts_before, faces_before))
viewer.dims.ndisplay = 3
viewer.reset_view()
```
image


```python
# displacement field is +1 mm along x, mesh and volume should still be aligned
verts_after, faces_after = resampled.get_mesh("SSp")
roi_mask_after = resampled.get_masks("SSp")
viewer, _ = roi_mask_after.fusi.plot()
layer = viewer.add_surface((verts_after, faces_after))

viewer.dims.ndisplay = 3
viewer.reset_view()

```

image

FelipeCybis:
```python
verts_after, faces_after = resampled.get_mesh("SSp")
roi_mask_after = resampled.get_masks("SSp")
viewer, _ = roi_mask_after.fusi.plot()
viewer, _ = roi_mask_before.fusi.plot(colormap="magenta", viewer=viewer)
layer = viewer.add_surface((verts_after, faces_after))

viewer.dims.ndisplay = 3
viewer.reset_view()
```

with the fix
image

FelipeCybis (approved):
Finally :P 
I will let you review my last commits, but this mega PR is all good for me!
Always a love-hate relationship with affine/bspline-related things!

PR #239, opened 2026-07-03
fix(signal): harden clean against boundary and non-finite inputs

### LINKED ISSUE #238: cf.signal.clean: boundary-censored timepoints produce NaN that propagates through temporal filter
## Description

When `sample_mask` is passed to `cf.signal.clean` together with a temporal filter (`low_cutoff` / `high_cutoff`), `clean` calls `interpolate_samples` internally to fill censored timepoints before filtering (the Lindquist et al. 2019 pre-scrubbing approach). However, `interpolate_samples` uses linear interpolation by default, which cannot extrapolate — so any censored timepoints **at the start or end** of the recording are left as `NaN`.

Those `NaN` values then propagate through `filter_butterworth`, making the entire filtered signal `NaN`, which causes `scipy.linalg.qr` to raise during confound regression:

```
ValueError: array must not contain infs or NaNs
```

## Minimal reproduction

```python
import numpy as np
import xarray as xr
import confusius as cf

t = np.arange(100, dtype=float)
signals = xr.DataArray(np.random.randn(100, 10), dims=["time", "space"], coords={"time": t})
confounds = xr.DataArray(np.random.randn(100, 5), dims=["time", "confound"], coords={"time": t})

mask_values = np.ones(100, dtype=bool)
mask_values[0] = False  # censor first timepoint
sample_mask = xr.DataArray(mask_values, dims=["time"], coords={"time": t})

# Raises: ValueError: array must not contain infs or NaNs
cf.signal.clean(signals, low_cutoff=0.01, confounds=confounds, sample_mask=sample_mask)
```

## Root cause

In `signal/core.py`, `clean` calls:

```python
signals = interpolate_samples(signals, sample_mask=sample_mask, method=interpolate_method)
```

with the default `interpolate_method="linear"`. `interpolate_samples` uses `xr.DataArray.interp`, which leaves boundary positions as `NaN` when it cannot interpolate (no valid neighbour on one side). The resulting `NaN` then propagates through `filter_butterworth`.

## Suggested fix

`interpolate_samples` should fall back to the nearest valid value for boundary positions, for example by forwarding `fill_value="extrapolate"` by default, or by applying `ffill`/`bfill` after interpolation. Alternatively, `clean` could expose an `interpolate_kwargs` parameter so callers can pass `fill_value="extrapolate"` themselves.

Note: the `interpolate_samples` docstring example already shows `fill_value="extrapolate"` as the workaround, but labels it `# Default: extrapolate at boundaries.` which is misleading — it is not the default.

sdiebolt:
## Summary
- add `ensure_finite` and `interpolate_kwargs` to `confusius.signal.clean`
- repair non-finite values with per-series time interpolation when `ensure_finite=True`, while leaving them untouched when `False`
- fill boundary-censored samples from the nearest kept sample before detrending/filtering
- fix the misleading `interpolate_samples` boundary doc example and add regression tests

Closes #238.

### DIFF
--- a/src/confusius/signal/core.py
+++ b/src/confusius/signal/core.py
@@ -19,7 +19,7 @@ def _interpolate_non_finite(data: xr.DataArray, *, name: str) -> xr.DataArray:
 
     Parameters
     ----------
-    data : xarray.DataArray
+    data : (time, ...) xarray.DataArray
         Data to sanitize.
     name : str
         Name used in error messages.
@@ -56,7 +56,7 @@ def _fill_interpolated_boundary_non_finite(
 
     Parameters
     ----------
-    signals : xarray.DataArray
+    signals : (time, ...) xarray.DataArray
         Interpolated signals.
     sample_mask : xarray.DataArray
         Boolean sample mask already validated by `interpolate_samples`.
--- a/src/confusius/signal/core.py
+++ b/src/confusius/signal/core.py
@@ -14,6 +14,91 @@ from confusius.signal.standardization import standardize
 from confusius.validation import validate_time_series
 
 
+def _interpolate_non_finite(data: xr.DataArray, *, name: str) -> xr.DataArray:
+    """Interpolate non-finite values along time and fill boundaries.
+
+    Parameters
+    ----------
+    data : xarray.DataArray
+        Data to sanitize.
+    name : str
+        Name used in error messages.
+
+    Returns
+    -------
+    xarray.DataArray
+        Data with non-finite values replaced by time interpolation, with leading and
+        trailing gaps filled from the nearest finite sample.
+
+    Raises
+    ------
+    ValueError
+        If any series remains entirely non-finite after interpolation.
+    """
+    finite = xr.apply_ufunc(np.isfinite, data, dask="allowed")
+    if not bool(finite.any(dim="time").all().compute().item()):
+        raise ValueError(
+            f"{name} contains a series with no finite values along time, so "
+            "ensure_finite=True cannot repair it."
+        )
+
+    repaired = data.where(finite)
+    repaired = repaired.interpolate_na(dim="time", method="linear")
+    repaired = repaired.ffill("time").bfill("time")
+
+    return repaired
+
+
+def _fill_interpolated_boundary_non_finite(
+    signals: xr.DataArray, sample_mask: xr.DataArray
+) -> xr.DataArray:
+    """Fill boundary non-finite values from the nearest kept sample.
+
+    Parameters
+    ----------
+    signals : xarray.DataArray
+        Interpolated signals.
+    sample_mask : xarray.DataArray
+        Boolean sample mask already validated by `interpolate_samples`.
+
+    Returns
+    -------
+    xarray.DataArray
+        Signals with any non-finite values left on censored boundaries replaced by the
+        nearest kept sample.
+    """
+    boolean_mask = np.asarray(sample_mask.values)
+    first_kept = int(np.argmax(boolean_mask))
+    last_kept = len(boolean_mask) - int(np.argmax(boolean_mask[::-1])) - 1
+    result = signals
+
+    if first_kept > 0:
+        left_boundary = xr.DataArray(
+            np.arange(signals.sizes["time"]) < first_kept,
+            dims=["time"],
+            coords={"time": signals.coords["time"]},
+        )
+        left_fill = signals.isel(time=first_kept).broadcast_like(signals)
+        left_missing = left_boundary & ~xr.apply_ufunc(
+            np.isfinite, result, dask="allowed"
+        )
+        result = xr.where(left_missing, left_fill, result)
+
+    if last_kept < signals.sizes["time"] - 1:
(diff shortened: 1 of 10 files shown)


sdiebolt:
@FelipeCybis ready to review!

src/confusius/signal/core.py
```diff
@@ -14,6 +14,91 @@
...
+            dims=["time"],
+            coords={"time": signals.coords["time"]},
+        )
+        right_fill = signals.isel(time=last_kept).broadcast_like(signals)
+        right_missing = right_boundary & ~xr.apply_ufunc(
+            np.isfinite, result, dask="allowed"
+        )
+        result = xr.where(right_missing, right_fill, result)
+
+    return result
```
FelipeCybis, line 99: Isn't this `ffill + bfill`? see my following comment below (and then this would not be needed)

src/confusius/signal/core.py
```diff
@@ -14,6 +14,91 @@
...
+
+
+def _fill_interpolated_boundary_non_finite(
+    signals: xr.DataArray, sample_mask: xr.DataArray
+) -> xr.DataArray:
+    """Fill boundary non-finite values from the nearest kept sample.
+
+    Parameters
+    ----------
+    signals : xarray.DataArray
```
FelipeCybis:
```suggestion
    signals : (time, ...) xarray.DataArray
```

src/confusius/signal/core.py
```diff
@@ -14,6 +14,91 @@
 from confusius.validation import validate_time_series
 
 
+def _interpolate_non_finite(data: xr.DataArray, *, name: str) -> xr.DataArray:
+    """Interpolate non-finite values along time and fill boundaries.
+
+    Parameters
+    ----------
+    data : xarray.DataArray
```
FelipeCybis:
```suggestion
    data : (time, ...) xarray.DataArray
```

src/confusius/signal/core.py
```diff
@@ -131,25 +226,48 @@ def clean(
...
     # applied to the full time series without gaps.
     if sample_mask is not None and (detrend_order is not None or do_filter):
         signals = interpolate_samples(
-            signals, sample_mask=sample_mask, method=interpolate_method
+            signals,
+            sample_mask=sample_mask,
+            method=interpolate_method,
+            **interpolate_kwargs,
         )
+        signals = _fill_interpolated_boundary_non_finite(signals, sample_mask)
```
FelipeCybis, line 262:
```suggestion
        signals = signals.ffill("time").bfill("time")
```
after `intermpolate_samples` this could be only `ffill`, `bfill` no?

src/confusius/signal/core.py
```diff
@@ -131,25 +226,48 @@ def clean(
...
         if confounds is not None:
             confounds = interpolate_samples(
-                confounds, sample_mask=sample_mask, method=interpolate_method
+                confounds,
+                sample_mask=sample_mask,
+                method=interpolate_method,
+                **interpolate_kwargs,
             )
+            confounds = _fill_interpolated_boundary_non_finite(confounds, sample_mask)
```
FelipeCybis, line 270:
```suggestion
            confounds = confounds.ffill("time").bfill("time")
```

after `interpolate_samples` this could be only `confounds = confounds.ffill("time").bfill("time")` no?

src/confusius/signal/core.py
sdiebolt: Nope, ffill + bfill would also replace non-finite values inside boundaries!

FelipeCybis: Yes, but after `interpolate_samples` there is no more non-finite values inside boundaries, right?

sdiebolt: There is. `interpolate_samples` should only interpolate on new coordinates, not interpolate non-finite values away.

sdiebolt: The goal of `ensure_finite=False` is to be a contract saying "I'm sure I have no non-finite values". If you do have non-finite values, you'll either have hard error (from QR decomposition for example), or a DataArray full of NaNs (from filtering), which is an obvious indication of a problem in the recording. Kind of like `skip_na` in NumPy/SciPy functions.

sdiebolt: (And also `interpolate_samples` doesn't necessarily run!)

FelipeCybis: I am not sure I am completely following ahah. We have a `sample_mask` that goes through `interpolate_samples`.  If there are bad samples at extremities, it returns as `NaN` with `linear` method and that is the root of the problem, right? But we could possibly have non-finite values in good samples of the sample mask and we don't want to fill those except if `ensure_finite`?

sdiebolt: Right on point! If there are non-finite values pre-existing in the recording and that were not removed by the sample mask, we need to the user to be aware, not silently remove them. Does that make sense?

FelipeCybis: Understood! All good!

FelipeCybis (approved):

PR #242, opened 2026-07-04
feat(plotting): add plot_stat_map convenience wrapper for statistical overlays

### LINKED ISSUE #240: feat: add plot_stat_map convenience wrapper for statistical overlays
## Summary

`plot_volume` + `add_volume` requires callers to manually wire up a recurring pattern for statistical maps: plot a background anatomical volume, overlay the stat map fully opaque, and pick a symmetric `vmin`/`vmax`. A `plot_stat_map` wrapper would encode these defaults once.

sdiebolt:
## Summary

- Adds `plot_stat_map(stat_map, bg_volume=None, ...)` and the matching `data.fusi.plot.stat_map` accessor: encodes the recurring `plot_volume` + `add_volume` pattern for statistical overlays (optional background, fully opaque overlay, symmetric colormap range defaulting to the 98th percentile of `|stat_map|`, `coolwarm` by default).
- `bg_kwargs` forwards cmap/vmin/vmax/norm/alpha overrides to the background layer without polluting the top-level signature.
- Fixes `just generate-baselines`, which previously wiped all baseline PNGs but only regenerated the volume/contours classes, silently deleting composite/carpet baselines.

Closes #240.


sdiebolt:
@FelipeCybis Ready to review!

docs/examples/04_connectivity/02_atlas_seed_map.py
```diff
@@ -0,0 +1,214 @@
...
+# [Allen Mouse Brain Atlas][confusius.atlas.Atlas] into the recording's native space,
+# pick four atlas regions of interest as seeds, and correlate each seed's signal
+# against every voxel with [`SeedBasedMaps`][confusius.connectivity.SeedBasedMaps]. Each
+# resulting map is displayed with
+# [`plot_stat_map`][confusius.plotting.plot_stat_map], using the resampled Allen
+# reference volume as background.
+#
+# We reuse the same recording, template, and registration workflow as the
+# atlas-based region correlation matrix example (see its own page for a detailed
+# walkthrough of the registration steps condensed here).
```
FelipeCybis: TODO: when #243 is merged

src/confusius/plotting/image.py
```diff
@@ -2195,6 +2326,293 @@ def plot_composite(
...
+            axes=axes,
+            bg_color=bg_color,
+            fg_color=fg_color,
+            yincrease=yincrease,
+            xincrease=xincrease,
+        )
+
+    resolved_vmin, resolved_vmax, resolved_cmap = _resolve_stat_map_style(
+        stat_map, vmin, vmax, cmap, auto_range
+    )
```
FelipeCybis, line 2521: The only thing that I am wondering about this is that the range will be computed from the whole data even if you want to plot a subset of it through `slice_coords`. But computing the slices here kind of defeats the purpose of just using `add_volume` with opinionated arguments. What do you think @sdiebolt ?

sdiebolt: Hmmm... That's a good observation. I'd say that even if you plot only a few slices, you'd want the range to be adjusted to the whole map generally? So that it stays constant even as you change the slice coordinates? Although we could surely find cases where you'd prefer it to be adjusted to only the sliced coords... But in that case I guess you can always use `.sel/.isel` before calling `plot_stat_map`? What do you think?

FelipeCybis: Yeah, I guess using `.sel/.isel` ok for solving this.

FelipeCybis (approved):
LGTM!

PR #243, opened 2026-07-04
feat(plotting): add plot_matrix for connectivity/correlation matrices

### LINKED ISSUE #241: feat: add plot_matrix for connectivity/correlation matrices
## Summary

We have no way to visualize 2D matrices (connectivity, correlation, region-by-region matrices) — only volumetric plots. A `plot_matrix` function, inspired by nilearn's `nilearn.plotting.plot_matrix`, would fill this gap. It should also support an optional "label groups" feature: colored rectangles drawn alongside row/column labels to mark groups of labels (e.g. cortex, thalamus, ...), useful when there are too many individual labels to read or when highlighting anatomical groupings. Nilearn's `plot_matrix` has no equivalent.

sdiebolt:
## Summary
- Adds `plot_matrix`, inspired by `nilearn.plotting.plot_matrix`, for plotting 2D matrices (`numpy.ndarray` or `xarray.DataArray`) with lower/diag triangle masking, grid lines, and a symmetric diverging colormap.
- Adds a `groups` feature (no Nilearn equivalent): contiguous label runs get colored rectangle strips alongside the matrix, useful for marking anatomical groupings (e.g. cortex, thalamus) when there are too many individual labels to read.
- Moves color/fontsize helpers shared between `image.py` and the new `matrix.py` into module-private `plotting/_utils.py`.
- Fixes `just generate-baselines`, which only regenerated two of the visual-regression test classes and was silently deleting the rest.

Closes #241


sdiebolt:
@FelipeCybis Ready to review!

docs/examples/04_connectivity/01_atlas_correlation_matrix.py
```diff
@@ -0,0 +1,275 @@
...
+        [0.0, 0.0, 1.0, 5.6650243788545875],
+        [0.0, 0.0, 0.0, 1.0],
+    ]
+)
+initialization = np.linalg.inv(napari_affine)
+
+# Crop the template to a thin band around the recording's expected location to improve
+# registration speed and visualization.
+target_z = napari_affine[0, 3] + float(moving.z.values[0])
+fixed = template.sel(z=slice(target_z - 1.0, target_z + 1.0)).fusi.scale.db()
```
FelipeCybis, line 92: Nice :)

docs/examples/04_connectivity/01_atlas_correlation_matrix.py
```diff
@@ -9,8 +9,8 @@
 #
 # We use an awake freely-running acquisition from subject `CR022`, session `20201007`,
 # in the [Nunez-Elizalde 2022 dataset][confusius.datasets.fetch_nunez_elizalde_2022],
-# and the [Pepe, Mariani 2026 fUSI
-# template][confusius.datasets.fetch_template_pepe_mariani_2026], which carries the
+# and the [Pepe, Mariani 2026 fUSI template][confusius.datasets.fetch_template_pepe_mariani_2026],
```
FelipeCybis, line 12: This would not create the link if wrapped in two lines.

FelipeCybis (approved):
Awesome! The examples in this PR and on the `plot_stat_map` PR are so cool!!

Made  some small changes that I will let you check.
I think for the most important one is using pcolormesh instead of imshow, which keeps, at any zoom, the grid lines and cell edges aligned.

sdiebolt:
> Awesome! The examples in this PR and on the `plot_stat_map` PR are so cool!!
> 
> Made some small changes that I will let you check. I think for the most important one is using pcolormesh instead of imshow, which keeps, at any zoom, the grid lines and cell edges aligned.

Indeed, we're progressively getting to the stats examples and it really shows how far we've pushed ConfUSIus :star_struck: 

Thanks for the fixes! Indeed, I didn't see at first the grid shift but it was clearly visible from the baseline images! 

Good to merge when green!

PR #245, opened 2026-07-05
fix(xarray): scale voxdim with applied affine

### LINKED ISSUE #244: `voxdim` should be updated to the current zoom of the of the applied transform

For NIfTI files, `voxdim` is build from the zoom of the primary affine (fallback to pixdim when no affine is valid).

Whenever `apply_affine` is used, `voxdim` should be updated to the new applied affine and corrrespond to the zoom of that affine.
This is especially needed for singleton dimensions since `data.fusi.spacing` will fallback to the dim's `voxdim`! This means that a NIfTI file round trip that passes through an `apply_affine` will save wrong affines if the data has a singleton spatial dimension.

FelipeCybis:
## Summary
- `apply_affine` now rescales each spatial coordinate's `voxdim` attribute by the absolute per-axis zoom, keeping the stored voxel size in sync with the rescaled coordinate values
- add a regression test asserting `voxdim` scales with the applied zoom (and that the input array is not mutated)
- add a NIfTI round-trip regression test: a file with sform + qform and a singleton spatial axis, loaded, transformed with `physical_to_qform`, and saved must keep its original qform — the singleton axis has no coordinate steps, so the saved spacing relies entirely on `voxdim`
- add changelog entry

Closes #244.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -57,6 +57,9 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- [`apply_affine`][confusius.xarray.apply_affine] now rescales the `voxdim` attribute
+  of the spatial coordinates along with the coordinate values
+  ([#245](https://github.com/confusius-tools/confusius/pull/245)).
 - `signal.clean` now supports `ensure_finite=True` to repair non-finite `signals`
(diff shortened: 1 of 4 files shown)


FelipeCybis:

sdiebolt (approved):
LGTM!

PR #249, opened 2026-07-05
fix(atlas,extract): disambiguate hemisphere masks by name, not raw id

### LINKED ISSUE #248: Atlas.get_masks + extract_with_labels don't compose for left/right hemisphere masks
## Problem

Building region correlation matrices with per-hemisphere masks requires manual workarounds today, discovered while writing the atlas correlation matrix example (`docs/examples/04_connectivity/01_atlas_correlation_matrix.py`):

1. **`Atlas.get_masks`'s `mask` coordinate ignores `side`.** Calling `atlas.get_masks(["VISp", "VISp"], sides=["left", "right"])` returns a `mask` coordinate of `["VISp", "VISp"]` — both layers also carry the same region id (`rid`). Callers have to reassign the `mask` coordinate themselves to disambiguate hemispheres.

2. **`extract_with_labels` rejects duplicate region ids across stacked mask layers**, even though each layer is already uniquely identified by its position along the `mask` dimension. Piping the output of (1) directly into `extract_with_labels` raises:

   ```
   ValueError: Stacked mask layers have duplicate non-zero IDs: {rid}. Each layer must carry a unique region ID.
   ```

   The example works around this by manually reassigning fresh ids to every mask layer with `xr.where` before calling `extract_with_labels`.

## Fix

- `Atlas.get_masks`: suffix the acronym with `_L`/`_R` when the corresponding `side` is `"left"`/`"right"` (bare acronym when `"both"`, the default), so same-region-different-side calls produce distinct `mask` coordinate values automatically.
- `extract_with_labels`: stop keying the flox reduction on the stacked mask layer's own non-zero value. Assign a synthetic per-layer id from the layer's position along `mask` instead, then map back to the `mask` coordinate for the output `region` names. This removes the duplicate-id restriction entirely — a layer's position already disambiguates it — and lets `Atlas.get_masks` output be passed straight through without manual relabeling.

Considered making `get_masks` return negative ids for the left hemisphere instead of a string suffix, but rejected it: `rid` also keys `annotation.attrs["rgb_lookup"]`, `roi_labels`, and the atlas cmap/norm, so negative ids would need `abs()`-ing at every one of those call sites for no remaining correctness benefit once the `mask`-coordinate fix above lands.

sdiebolt:
## Summary
- `Atlas.get_masks` now suffixes the `mask` coord acronym with `_L`/`_R` for `sides="left"`/`"right"` (bare acronym for `"both"`), so requesting the same region on both hemispheres no longer produces duplicate `mask` coord values.
- `extract_with_labels` no longer keys its flox reduction on a stacked mask layer's own non-zero value — it assigns a synthetic per-layer id from position along `mask` instead, so `get_masks` output with duplicate region ids across layers can be passed straight through without manual relabeling.

Closes #248.

## Test plan
- [x] `uv run pytest tests/unit/test_extract/test_labels.py tests/unit/test_atlas/test_atlas.py -q` (71 passed, includes new duplicate-id and side-suffix regression tests)
- [x] `just pre-commit` (ruff, ty, codespell, numpydoc-validation)

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -62,13 +62,12 @@ Current development version for the next ConfUSIus release.
 ### :bug: Fixes
 
 - [`Atlas.get_masks`][confusius.atlas.Atlas.get_masks] now suffixes the `mask`
-  coordinate with `_L`/`_R` for `sides="left"`/`"right"`, so requesting the same
-  region on both hemispheres no longer produces duplicate `mask` values.
+  coordinate with `_L`/`_R` for `sides="left"`/`"right"`, so requesting the same region
+  on both hemispheres no longer produces duplicate `mask` values.
   [`extract_with_labels`][confusius.extract.extract_with_labels] no longer requires
-  unique region ids across stacked mask layers — a layer is already identified by its
-  position along `mask` — so `get_masks` output can be passed straight through
-  without manual relabeling
-  ([#249](https://github.com/confusius-tools/confusius/pull/249)).
+  unique region ids across stacked mask layers—a layer is already identified by its
+  position along `mask`—so `get_masks` output can be passed straight through without
+  manual relabeling ([#249](https://github.com/confusius-tools/confusius/pull/249)).
 - [`apply_affine`][confusius.xarray.apply_affine] now rescales the `voxdim` attribute
   of the spatial coordinates along with the coordinate values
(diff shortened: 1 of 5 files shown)


sdiebolt:
@FelipeCybis ready to review!

FelipeCybis (approved):
LGTM

PR #251, opened 2026-07-06
fix(plotting): support non-numeric slice_mode coordinates

### LINKED ISSUE #250: Support slicing along non-numeric coordinates (e.g. region/mask labels) in plotting functions
`plot_volume`, `plot_stat_map`, `add_contours`, etc. slice via a `slice_mode` dimension name (e.g. `"x"`, `"y"`, `"z"`, `"time"`), but the underlying machinery assumes that dimension has numeric coordinates.

Panel selection uses nearest-neighbour lookup, `data.sel({slice_mode: coord}, method="nearest")`, which requires a numeric index and raises on a string-labeled coordinate:

```pycon
>>> data.sel(region="b", method="nearest")
TypeError: operation 'sub' not supported for dtype 'str' with dtype 'str'
```

Panel titles are built as `f"{slice_mode} = {coord:.3g} {units}"` (`VolumePlotter._build_slice_title`), which raises a `ValueError` on a non-numeric `coord`.

This came up trying to plot a `(region, z, y, x)` seed-based connectivity map (`SeedBasedMaps.maps_`, one Pearson-r map per named seed region) with a single `plot_stat_map(..., slice_mode="region")` call instead of looping over `.sel(region=...)` per panel — see the loop-based workaround in `docs/examples/04_connectivity/02_atlas_seed_map.py`.

sdiebolt:
## Summary
- `VolumePlotter` (and the `plot_volume`/`plot_contours`/`plot_composite` wrappers) always used `method="nearest"` for coordinate lookup and `.3g`-formatted panel titles/warnings, both of which crash on non-numeric `slice_mode` coordinates (e.g. a `region` label dimension from `SeedBasedMaps.maps_`).
- Numeric coordinates still use nearest-neighbour lookup; non-numeric coordinates now use exact matching, and titles/warnings format non-numeric values as plain strings instead of `.3g`.

Closes #250.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -58,6 +58,14 @@ Current development version for the next ConfUSIus release.
   `data.fusi.affine.apply` accessor now accept a string naming a key in
   `attrs["affines"]`, instead of requiring the affine matrix itself
   ([#247](https://github.com/confusius-tools/confusius/pull/247)).
+- Plotting functions that slice along `slice_mode` (
+  [`plot_volume`][confusius.plotting.plot_volume],
+  [`plot_contours`][confusius.plotting.plot_contours],
+  [`plot_composite`][confusius.plotting.plot_composite], and the
+  [`VolumePlotter`][confusius.plotting.VolumePlotter] methods) now support non-numeric
+  coordinates (e.g. region/mask labels), so a single call can slice a stacked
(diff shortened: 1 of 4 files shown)


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (approved):
LGTM

PR #253, opened 2026-07-06
refactor(nifti): simplify pixdim writing on save

### LINKED ISSUE #252: Remove NIfTI header information that is overwritten on save
On save of NIfTI file we set zooms, then we set pixdim, then we set qform.

https://github.com/confusius-tools/confusius/blob/00133de125972bdd273c0bbb66ac1b1cfc56cefc/src/confusius/io/nifti.py#L2028-L2034

Looking at `nibabel` code. [Setting `zooms`](https://github.com/nipy/nibabel/blob/4704d904b1bd50edb4c2429db1ab8cecd3e4ec2a/nibabel/analyze.py#L692), actually sets the `pixdim`. and [setting the `qform`](https://github.com/nipy/nibabel/blob/4704d904b1bd50edb4c2429db1ab8cecd3e4ec2a/nibabel/nifti1.py#L2075) also sets the `pixdim`. In the end, setting the `qform` just makes the lines above useless. I propose, for the sake of clarity, we remove them adding a comment that the `qform` sets the `pixdim` in the NIfTI header (we always have a valid `qform` here).

FelipeCybis:
- Closes #252 

Simplifies how `save_nifti` writes voxel spacings into the NIfTI header `pixdim`.

Previously `_create_nifti_image` called nibabel's `set_zooms` (with absolute values, then a loop to restore signs) before `set_qform`. But `set_qform` rewrites the spatial zooms (`pixdim[1:4]`) and `qfac` (`pixdim[0]`), so the `set_zooms` spatial write was immediately overwritten and wasted.

Now only the axes `set_qform` never touches are written, directly into `pixdim[4:]`:

- `pixdim[4]` — temporal spacing (`tr_pixdim`)
- `pixdim[5:8]` — extra-dimension spacings (`dim4`/`dim5`/`dim6`)

Signs are preserved by writing the values directly (nibabel's `set_zooms` forces zooms positive), which the extra-dim loader relies on to reconstruct coordinates from a signed step.

## Behavior

No user-facing change. Verified that the produced header is **byte-identical** to the previous implementation across cases with negative spatial spacing, a temporal axis, and a negatively-spaced extra dimension.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -74,13 +74,21 @@ Current development version for the next ConfUSIus release.
   the coord does not start at 0 with regular spacing); otherwise the spacing is stored
   in `pixdim` and the coord is rebuilt as `step * arange(size)` on load. Attributes are
   preserved in `ConfUSIusDim{N}Attributes` entries.
-  ([#223](htttps://github.com/confusius-tools/confusius/pull/223)).
+  ([#223](https://github.com/confusius-tools/confusius/pull/223)).
 
 ### :books: Documentation
 
 - Add an [NMF example](examples/_built/decomposition/nmf_single_recording.md) to the
   gallery, demonstrating the z-score + absolute-value standardization that makes
   signed fUSI signals NMF-compatible.
+
+### :wrench: Maintenance
+
+- Simplified the NIfTI save path: time and extra-dimension voxel spacings are now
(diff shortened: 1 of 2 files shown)


FelipeCybis:

sdiebolt (approved):
LGTM!

PR #255, opened 2026-07-06
fix(nifti,napari): drop non-serializable sidecar attrs, fall back to gray colormap

### LINKED ISSUE #254: NIfTI saver serializes non-JSON-serializable attrs as str() instead of skipping them
`save_nifti` writes the JSON sidecar with `json.dump(..., default=str)` (nifti.py line 2206), silently converting any non-serializable attr to its `str()` repr instead of skipping it.

Non-serializable attrs such as `ListedColormap` and `BoundaryNorm` (present in `Atlas.annotation.attrs`) get written as e.g. `"<matplotlib.colors.ListedColormap object at 0x...>"`. On reload that string lands in `attrs["cmap"]` and napari crashes with a `KeyError` when trying to use it as a colormap name.

The zarr backend already skips these correctly — the NIfTI path should do the same.

**Fix:** pre-filter attrs with `json.dumps` (no `default=str`) before building the sidecar, and remove `default=str` from the final `json.dump` call.

sdiebolt:
## Summary
- `save_nifti` now drops attrs that aren't JSON-serializable as-is instead of stringifying them via `default=str`, which could corrupt fields like `cmap` on reload.
- The napari zarr/NIfTI reader now falls back to the `gray` colormap (with a warning) instead of crashing when a layer's `cmap` attr isn't a valid napari colormap.

Fixes #254.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -72,9 +72,11 @@ Current development version for the next ConfUSIus release.
 - `save_nifti` now drops attrs that cannot be serialized to JSON as-is (e.g. matplotlib
   `ListedColormap`/`BoundaryNorm` objects) instead of writing their `str()` repr into the
   sidecar, which could corrupt fields such as `cmap` on reload. A warning lists the
-  dropped keys. **[Napari plugin]** The reader now falls back to the `"gray"` colormap
-  (with a napari warning) instead of crashing when a layer's `cmap` attr is not a valid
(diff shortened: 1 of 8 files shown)


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (approved):
LGTM

PR #261, opened 2026-07-07
feat(datasets): re-download files with changed upstream md5 on refresh

### LINKED ISSUE #260: Dataset refresh should re-download files whose upstream md5 changed
## Problem

The dataset fetchers (`fetch_cybis_pereira_2026`, `fetch_nunez_elizalde_2022`, `fetch_landemard_2026`) expose a `refresh` flag, but it only decides whether the *index* is re-fetched and whether *missing* files are downloaded. It cannot detect that a file already on disk has **changed upstream**:

- `download_missing_osf_files` (`src/confusius/datasets/_osf.py`) selected files with `not (bids_dir / p).exists()` — a file that exists was never reconsidered.
- `retrieve_with_retries` (`src/confusius/datasets/_pooch.py`) called `pooch.retrieve(..., known_hash=None)`, so downloads were never integrity-checked either.

So if a dataset file is corrected/reprocessed and re-uploaded, `refresh=True` re-fetched the index and picked up *newly added* paths, but silently kept the stale local copy of any path that already existed.

## What changed upstream

The per-dataset upload scripts now write an `md5` field into `dataset_index.json`:

```json
{"osf_path": "/<id>", "size": 132, "md5": "<hex md5>"}
```

Going forward the remote index always carries `md5` (the upload scripts also backfill it into older indices). Those upload/convert repos are separate from confusius, so nothing in confusius should depend on maintainers re-running them.

FelipeCybis:
Closes #260.

## Problem

The dataset fetchers' `refresh` flag only decided whether the *index* was re-fetched and whether *missing* files were downloaded. A file already on disk was never reconsidered (`download_missing_osf_files` selected `not (bids_dir / p).exists()`), and downloads used `known_hash=None`, so an upstream content change was silently ignored on `refresh=True`.

## Change

The dataset index now carries an `md5` per file. Rather than re-hashing local files (a full re-hash of a multi-GB dataset on every refresh is not acceptable), this trusts the cached `dataset_index.json` as the record of what is on disk and compares **index-to-index**.

On `refresh=True`, the fetcher captures the cached index *before* it is touched, fetches the remote index, and for each requested file re-downloads when the file is **missing** from disk, **changed** (remote md5 differs from the cached md5), or **unverifiable** (its cached entry has no usable md5 — a null hash, or a path not previously catalogued). Downloads pass `known_hash="md5:<hash>"` to pooch, so every download (refresh or first fetch) is integrity-verified on arrival. `refresh=False` is unchanged: download-missing-only, offline-friendly.

### Merge, don't replace, the cached index

`get_index` no longer overwrites the cache on refresh. After downloads, `update_cached_index` persists a **merged** index: the remote entry for every reconciled (requested) file, the previously cached entry for any other known file, and the remote entry for files never seen before. So a *filtered* refresh (e.g. `subjects=[...]`) updates only the files it actually reconciled and keeps the md5 baseline of everything else — an unrequested file that changed upstream is still detected the next time it is refreshed, rather than being silently marked current.

### Validate the cached index schema

We now assume every dataset index carries md5 going forward and stop degrading gracefully for older layouts. Whenever a locally cached `dataset_index.json` is read, its structure is validated: every entry must define `osf_path`, `size`, and `md5`. If it doesn't (e.g. a dataset downloaded by an older confusius release), the fetch fails fast with a clear error that names the dataset directory to delete and re-fetch. The check is structural, so any future index-schema change is caught the same way rather than silently mishandled. `OsfFileInfo.md5` is now a required (nullable) key.

The dataset upload/convert repos are separate from confusius, so nothing here asks maintainers to re-run them; the remote index always carries md5 going forward.

## Files

- `_osf.py` — adds `read_cached_index` (now schema-validating), `update_cached_index`, `_validate_index`, `_INDEX_ENTRY_KEYS`, and a `_write_index` helper; `get_index` reads the cache through `read_cached_index` and persists only on first fetch; `download_missing_osf_files` → `download_osf_files(bids_dir, files, previous_index=None, refresh=False)` selects missing/changed/unverifiable files via `_select_downloads` and threads md5 into `known_hash`. `OsfFileInfo.md5` becomes `str | None`.
- `_pooch.py` — `retrieve_with_retries` accepts `known_hash` and forwards it to `pooch.retrieve`.
- `_nunez_elizalde_2022.py`, `_cybis_pereira_2026.py`, `_landemard_2026.py` — read the cached index before refresh, pass it through, and merge-persist afterwards; `refresh` docstrings updated.
- The template fetcher (`_huang_2025.py`) is unchanged: it fetches by direct URL, not via the index, and already deletes-then-redownloads on refresh.

## Tests

`test_osf.py` covers `read_cached_index` (absent → `{}`, valid → decoded, outdated structure → error naming the directory), `get_index` not persisting on refresh, `update_cached_index` (requested → remote, untouched → cached baseline, unseen → catalogued), and `download_osf_files`: cached file kept without refresh, refresh re-downloads on md5 change, refresh skips on unchanged md5, refresh re-downloads a file whose cached md5 is null, and known_hash forwarding on first download.

### DIFF
--- a/src/confusius/datasets/_osf.py
+++ b/src/confusius/datasets/_osf.py
@@ -233,7 +233,7 @@ def _select_downloads(
 
     A file is selected when it is absent from `bids_dir`. When `refresh` is
     `True`, a cached file is also selected when its md5 in `previous_index` (the
-    index it was last downloaded with) differs from — or is missing next to — its
+    index it was last downloaded with) differs from—or is missing next to—its
     md5 in `files` (the freshly fetched remote index). The cached index is trusted
     as the record of what is on disk, so files are never re-hashed; a cache entry
     that lacks an md5 (e.g. one predating md5 tracking) cannot be vouched for, so
--- a/src/confusius/datasets/_osf.py
+++ b/src/confusius/datasets/_osf.py
@@ -173,7 +173,7 @@ def update_cached_index(
     """Persist a refreshed index by merging, not replacing, the cached one.
 
     The cached `dataset_index.json` is the record of what is on disk, so it is
-    not overwritten wholesale with the remote index. Instead the merged index
+    not overwritten with the remote index. Instead the merged index
     is, per file:
 
     - the remote entry for every requested file (`files`), which has just been
--- a/src/confusius/datasets/_osf.py
+++ b/src/confusius/datasets/_osf.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import json
 from pathlib import Path
-from typing import NotRequired, TypedDict
+from typing import TypedDict
 
 import pooch
 import requests
@@ -27,18 +27,26 @@ _OSF_DOWNLOAD_BASE = "https://osf.io/download/{}/"
 _INDEX_FILENAME = "dataset_index.json"
 """Filename of the per-dataset index mapping BIDS-relative paths to metadata."""
 
+_INDEX_ENTRY_KEYS = frozenset({"osf_path", "size", "md5"})
+"""Keys every entry in a current `dataset_index.json` must define.
+
+A cached index whose entries do not all define these keys is from an older,
+incompatible confusius release and is rejected by
+[`read_cached_index`][confusius.datasets._osf.read_cached_index]."""
+
 
 class OsfFileInfo(TypedDict):
     """Per-file metadata entry in an OSF-backed dataset index.
 
-    Indices written before md5 tracking omit the `md5` key entirely; newer
-    ones may still carry `md5: null` for a file whose hash is unknown. Both
-    cases read back as a missing hash via `dict.get("md5")`.
+    Every current index defines `md5` for each file (its value may be `null`
+    for a file whose hash is unknown). A locally cached index missing the key
+    is from an older release and is rejected on load, so consumers can assume
+    the key is present.
     """
 
     osf_path: str
     size: int
-    md5: NotRequired[str | None]
+    md5: str | None
 
 
 def resolve_index_url(project_id: str, bids_root: str) -> str:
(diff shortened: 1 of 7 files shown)


FelipeCybis:
@sdiebolt This is good to be reviewed!

src/confusius/datasets/_osf.py
```diff
@@ -93,9 +99,16 @@ def get_index(
...
+    When `refresh` is `False` and a cached index exists in `data_dir`, it is
+    decoded and returned directly (offline-friendly). Otherwise the latest index
+    is fetched from OSF and returned.
+
+    The freshly fetched index is only persisted to disk on the very first fetch
+    (when no cache exists yet). On a refresh of an existing cache it is *not*
+    written back: the caller reconciles it against the cached index and persists
+    the merged result via
+    [`update_cached_index`][confusius.datasets._osf.update_cached_index], so that
+    files the caller did not reconsider keep their recorded md5.
```
sdiebolt:
```suggestion
    files the caller did not update keep their recorded md5.
```

src/confusius/datasets/_osf.py
```diff
@@ -115,40 +128,188 @@ def get_index(
...
+def update_cached_index(
+    data_dir: Path,
+    remote_index: dict[str, OsfFileInfo],
+    previous_index: dict[str, OsfFileInfo],
+    files: dict[str, OsfFileInfo],
+) -> None:
+    """Persist a refreshed index by merging, not replacing, the cached one.
+
+    The cached `dataset_index.json` is the record of what is on disk, so it is
+    not overwritten wholesale with the remote index. Instead the merged index
```
sdiebolt:
```suggestion
    not overwritten with the remote index. Instead the merged index
```

src/confusius/datasets/_osf.py
```diff
@@ -115,40 +128,188 @@ def get_index(
...
+    is, per file:
+
+    - the remote entry for every requested file (`files`), which has just been
+      reconciled with disk (downloaded, or confirmed unchanged);
+    - the previously cached entry for any other file already known locally, so
+      an md5 baseline the caller did not reconsider is preserved rather than
+      silently advanced to the remote md5;
+    - otherwise the remote entry, cataloguing files not seen before.
```
sdiebolt: Wording here is weird. Terms like "reconsider" is a bit weird (but that might be me?). Also, I feel like the first two bullet points are overlapping? If the file is unchanged, then both points apply, although I feel like only the first one should? Maybe wording should be something like:

- the remote entry for every newly added or updated file.
- the local entry for any unchanged file, so that future downloads can still detect a mismatch.

src/confusius/datasets/_osf.py
```diff
@@ -115,40 +128,188 @@ def get_index(
...
+    bids_dir: Path,
+    files: dict[str, OsfFileInfo],
+    previous_index: dict[str, OsfFileInfo],
+    refresh: bool,
+) -> dict[str, OsfFileInfo]:
+    """Return the subset of `files` that must be (re)downloaded.
+
+    A file is selected when it is absent from `bids_dir`. When `refresh` is
+    `True`, a cached file is also selected when its md5 in `previous_index` (the
+    index it was last downloaded with) differs from — or is missing next to — its
```
sdiebolt:
```suggestion
    index it was last downloaded with) differs from—or is missing next to—its
```

src/confusius/datasets/_osf.py
FelipeCybis: Yes, the wording is different if we are talking about a _merged_ index or an _updated_ index.
I will work on this based on your suggestion.

FelipeCybis: reworked in [3f510f4](https://github.com/confusius-tools/confusius/pull/261/commits/3f510f4873cf47ef9010507d76866c7edd7695f2)

sdiebolt (approved):
LGTM!

PR #268, opened 2026-07-08
fix(plotting): stop sort_coords_for_plot from reordering slice_mode panels

### LINKED ISSUE #267: plot_stat_map/add_contours silently reorder panels when slice_mode coord isn't sorted
## Bug

`VolumePlotter._prepare_slice_inputs` (`image.py:654`) and `add_contours` (`image.py:1489`) both call `sort_coords_for_plot(data, data.dims)`, passing **all three dims** including `slice_mode` itself. `sort_coords_for_plot` sorts any non-monotonic-increasing coordinate — so if `slice_mode`'s own coordinate isn't already sorted (e.g. a `region` dim built from an arbitrary list of acronyms, or a descending `z`), panels get silently reordered.

The intended purpose of the sort (per its docstring) is to fix geometry for `pcolormesh`/contour/napari rendering *within* a panel — that only concerns the two display (row/col) dims, never the dim used to pick *which* slice becomes *which* panel.

## Repro

`docs/examples/04_connectivity/02_atlas_seed_map.py` calls `plot_stat_map(..., slice_mode="region")` with regions `["SSp-bfd", "RSP", "HIP", "VPM"]`. Since that list isn't alphabetically sorted, `sortby("region")` reorders panels to `HIP, RSP, SSp-bfd, VPM`, but manually-set titles (in the original notebook version) followed the input order — producing swapped `SSp-bfd`/`HIP` labels.

## Fix

Scope `sort_coords_for_plot` to the display dims only (exclude `slice_mode`) at both call sites.

sdiebolt:
## Summary
- `sort_coords_for_plot` was sorting all of `data.dims`, including `slice_mode` itself; only the two display dims need sorting for pcolormesh/contour geometry. Sorting `slice_mode` silently reorders which slice becomes which panel whenever that coordinate isn't already monotonic increasing (e.g. a `region` dim from an arbitrary acronym list, or descending `z`).
- Scoped the sort to display dims at both call sites: `VolumePlotter._prepare_slice_inputs` and `add_contours`.
- Updated the seed-map example to rely on `plot_stat_map`'s own titles instead of a manually zipped region list — that mismatch is what surfaced the bug (`SSp-bfd`/`HIP` labels were swapped).

Fixes #267.

## Test plan
- [x] Added regression test asserting panel titles follow the given (non-alphabetical) region order
- [x] `uv run pytest tests/unit/test_plotting/` — 178 passed
- [x] `uv run just pre-commit` — all hooks pass

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,7 +12,9 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
-- [`plot_stat_map`][confusius.plotting.plot_stat_map] and
+- [`plot_volume`][confusius.plotting.plot_volume],
+  [`plot_stat_map`][confusius.plotting.plot_stat_map],
+  [`plot_composite`][confusius.plotting.plot_composite], and
   [`VolumePlotter.add_contours`][confusius.plotting.VolumePlotter.add_contours] no
   longer silently reorder panels when `slice_mode`'s own coordinate isn't already
   sorted (e.g. a `region` dimension built from an arbitrary list of acronyms, or a
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -10,6 +10,15 @@ icon: lucide/history
 
 Current development version for the next ConfUSIus release.
 
+### :bug: Fixes
+
+- [`plot_stat_map`][confusius.plotting.plot_stat_map] and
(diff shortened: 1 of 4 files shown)


sdiebolt:
@FelipeCybis ready to review!

FelipeCybis (approved):
I ❤️ small PRs

PR #271, opened 2026-07-08
fix(plotting): suppress deprecated napari ScaleBar.unit warning

### LINKED ISSUE #270: plot_napari emits napari FutureWarning for ScaleBar.unit
\`plot_napari\` sets \`viewer.scale_bar.unit\` directly to show physical units on the scale bar, which napari 0.7.0 deprecated in favor of computing units from \`Layer.units\`. That replacement isn't implemented yet in napari (planned for 0.8.0/0.9.0), so there's no functional alternative today — the warning should just be suppressed at the source until napari catches up.

```
napari/plotting/napari.py:253: FutureWarning: Setting unit on the ScaleBar model is deprecated. Units will instead be computed from the layers in the layerlist. ...
  viewer.scale_bar.unit = scale_bar_unit
```

sdiebolt:
Suppresses the napari 0.7.0 FutureWarning from `plot_napari` when setting the scale bar unit. No functional replacement exists in napari 0.7.0 yet (see TODO comment).

Fixes #270

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,9 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- [`plot_napari`][confusius.plotting.plot_napari] no longer emits napari's
+  `FutureWarning` about the deprecated `ScaleBar.unit` API when `show_scale_bar=True`
+  ([#271](https://github.com/confusius-tools/confusius/pull/271)).
 - [`plot_volume`][confusius.plotting.plot_volume],
   [`plot_stat_map`][confusius.plotting.plot_stat_map],
   [`plot_composite`][confusius.plotting.plot_composite], and
--- a/src/confusius/plotting/napari.py
+++ b/src/confusius/plotting/napari.py
@@ -250,7 +250,19 @@ def plot_napari(
             (u for d, u in zip(all_dims, all_units) if d != time_dim and u is not None),
             "mm",
         )
-        viewer.scale_bar.unit = scale_bar_unit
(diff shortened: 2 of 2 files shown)


sdiebolt:
@FelipeCybis ready to review (another very short PR for a change :stuck_out_tongue:)

FelipeCybis (approved):

TimMonko:
Hi @sdiebolt, can you describe what is not sufficient/working for you with scale bar and unit inference? napari has implemented unit inference from the layers in 0.7.1 (when this warning was introduced). You can read more in this entire guide, but this particular section should hopefully provide more info: https://napari.org/stable/guides/units.html#how-the-scale-bar-gets-its-units

sdiebolt:
Hi @TimMonko, everything's working well! Actually the comment I added is wrong and I misunderstood what 0.7.1 was doing because I failed to properly update my environment (forgot to `uv lock --upgrade` and was still running 0.7.0, so still getting the warning). But we do need to ignore the warning in ConfUSIus as we're still supporting napari 0.7.0 and thus need to set the scale bar using `viewer.scale_bar.units = "mm"`. I'll probably open a new PR to bump napari to >=0.7.1 though, and remove this bit!

Thanks very much for the heads up, I missed the unit inference bit when checking the napari changelog!

P.S. By the way, I think the [scale bar example](https://napari.org/stable/gallery/scale_bar.html#sphx-glr-gallery-scale-bar-py) might need updating?

TimMonko:
Thanks for the reply! 
Yes the scale bar example did need updated, but it now has been for the upcoming release, you can see that at the dev docs https://napari.org/dev/gallery/scale_bar.html -- thanks for looking at it though!

PR #273, opened 2026-07-08
feat(napari): add dataset-backed sample loader

### LINKED ISSUE #272: feat(napari): add a ConfUSIus sample entry backed by dataset fetchers
Add a `File > Open Sample` entry for the ConfUSIus napari plugin that exposes a `Nunez-Elizalde 2022` sample, fetches it on demand, shows download progress with an abort button, and only downloads the minimum raw fUSI files needed for that recording.

sdiebolt:
## Summary
- add ConfUSIus napari sample entries for an awake 2D+t recording (Nunez-Elizalde 2022) and a pair of 2D slices (Cybis Pereira 2026).
- fetch only the selected raw fUSI recording and show download progress with abort support
- add dataset/datatype filters to the Nunez-Elizalde fetcher so the sample download stays minimal

Closes #272


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (approved):
LGTM. Added a single shot event to update the axis labels after the sample is loaded, so the labels are updated `(-4, -3, -2, -1) -> (time, z, y, x)`.

FelipeCybis:
My bad, skipped the last tests. fixing it...

Edit: this was interesting, the fake viewer we use didn't fire the timer I added during the test. but another test testing dimensions would fire it and cause the error. Running only `pytest tests/unit/test_napari/test_sample.py` would never cause the error because no other test would fire the timer event, but running the whole test set would.

PR #274, opened 2026-07-08
refactor(atlas): use xarray.Dataset as main atlas class + atlas xarray accessor

FelipeCybis:
## Atlas → Dataset

- Replaces the monolithic `Atlas` class with a plain `xarray.Dataset` plus a registered `.atlas` accessor for all atlas-aware operations.
- The Dataset carries `reference`, `annotation`, and `hemispheres` as data variables on a shared `(z, y, x)` grid.
- `fetch_brainglobe_atlas(name)` fetches any BrainGlobe atlas and returns the Dataset.
- The structure hierarchy rides in `attrs["structures"]` as the BrainGlobe `StructuresDict` itself, (de)serialized to JSON on load/save.
- The mesh transform is a single top-level `attrs["physical_to_base"]` pull transform (numpy affine or displacement-field DataArray), so `ds.physical_to_base` works uniformly and meshes stay aligned after any resample (renamed from `mesh_vertex_transform`).
- `get_mesh`, `get_masks`, and `search` are exposed both as accessor methods (`ds.atlas.get_mesh(...)`) and as standalone functions (`cf.atlas.get_mesh(ds, ...)`) that validate the Dataset first.
- `confusius.io` functions `save_atlas` / `load_atlas`: Zarr bundles arrays, structure hierarchy, and region meshes, so a reloaded atlas is immediately usable.

## Dataset validation

- `validate_atlas_dataset(ds, *, require_mesh_use=False)` enforces the schema every accessor and function relies on.
- Data variables: `reference`, `annotation`, and `hemispheres` are all present as data variables.
- Grid: the three variables share identical dims, and those dims are a subset of `(z, y, x)`, so a 2D resampled single slice is accepted.
- dtypes: `reference` is floating-point; `annotation` and `hemispheres` are integer-valued.
- Affines: where two variables both define an affine of the same name, the matrices must agree, guaranteeing a common physical frame.
- Mesh use (`require_mesh_use=True`, used by `get_mesh`): `attrs["physical_to_base"]` is present and at least one structure references a mesh file that exists on disk.

## Example

- New "Save and reload a resampled atlas" gallery example (`docs/examples/05_atlases_and_templates/01_saving_resampled_atlas.py`): register a recording, resample the Allen atlas onto its grid, `save_atlas` to Zarr, `load_atlas` back, and overlay annotations to confirm alignment.
- Adds a `collapse` / `collapse: <title>` gallery cell tag that hides setup/boilerplate behind a collapsed callout (used here to fold the fetch + register cell).

## User guide

- Rewrites `docs/user-guide/atlas.md` from "coming soon" into a full guide: fetching BrainGlobe atlases (plus a pointer to authoring your own), exploring the structure tree, `lookup` / `search`, plotting annotation contours, region meshes in napari, masks, and resample/save.
- Adds `docs/images/atlas/generate.py` for the annotation-contour and root-mesh figures, wired into `just generate-doc-images` and the docs CI workflow.

## Where to start reviewing

- `src/confusius/atlas/_accessor.py`: the `.atlas` accessor and the standalone `get_mesh` / `get_masks` / `search`, the core API surface.
- `src/confusius/validation/atlas.py`: the schema contract everything else assumes.
- `src/confusius/io/atlas.py`: `save_atlas` / `load_atlas`, including the StructuresDict↔JSON conversion and the DataArray-transform↔data-var bridge for Zarr.
- `src/confusius/atlas/_build.py` and `datasets/_brainglobe.py`: how a Dataset is constructed from BrainGlobe.

Important points:

- `physical_to_base` is the pull transform (physical → base), stored directly and never inverted on read; renamed from `mesh_vertex_transform`.
- `structures`, `physical_to_base` are present in the `Dataset` attributes and they are much of the core of what the atlas is. We do have a `validate_atlas_dataset` but still, that is a fragile point.
- Saving to zarr with metadata gives us this warning `ZarrUserWarning: Consolidated metadata is currently not part in the Zarr format 3 specification. It may not be supported by other zarr implementations and may change in the future.`. Can we ignore this?


FelipeCybis:
@sdiebolt this is good to be reviewed! big refactoring, but seems pretty robust by now. 

I do feel a lot better having this as `Dataset` and fUSI data as `DataArray` in terms of consistency. Immediate gain on refactoring custom `Atlas` to `Dataset` is marginal to be honest, but I think we will be able to do cool things as keeping statistical results and atlases in the same `.zarr` bundle.

sdiebolt (approved):
LGTM! This is a super nice refactoring that makes ConfUSIus even more interoperable with the broader scientific ecosystem! :rocket:

docs/examples/05_connectivity/02_atlas_seed_map.py
```diff
@@ -126,13 +119,15 @@
...
-# [`Atlas.get_masks`][confusius.atlas.Atlas.get_masks] returns a stacked
+# [`get_masks`][confusius.atlas.AtlasAccessor.get_masks] returns a stacked
 # `(mask, z, y, x)` integer DataArray — one layer per requested region — which
 # [`SeedBasedMaps`][confusius.connectivity.SeedBasedMaps] accepts directly as
 # `seed_masks`.
 
 # %%
-seed_masks = atlas_native.get_masks(["SSp-bfd", "RSP", "HIP", "VPM"], sides="right")
+seed_masks = atlas_native.atlas.get_masks(
+    ["SSp-bfd", "RSP", "HIP", "VPM"], sides="left"
```
sdiebolt: seed masks sides changed here during the merge I think!

sdiebolt: We should keep `right`, especially for the HIP map that looks better with right seed (recording has slightly lower signal in left DG I think).

FelipeCybis: ah, ok, the text still points to "left", though! I will change the text as well.

PR #278, opened 2026-07-09
refactor(xarray)!: introduce the VoxelData model

### LINKED ISSUE #198: Explore CoordinateTransformIndex-backed geometry
## Goal

Explore `CoordinateTransformIndex` (CTI) as a geometry model for voxel-affine data.

The main idea is to encode full physical position directly in coordinates. With plain 1D coordinates, rotations and shears cannot be represented, so they currently have to live in `attrs["affines"]`.

## Discussion points

- Can CTI replace the current coordinate/affine split?
- What should happen for oblique voxel data?
- How should this interact with existing xarray / serialization constraints?

sdiebolt:
Closes #198

## Summary

Introduces the voxel-to-world index: fUSI/VoxelData geometry is now stored as native voxel-space dims (`k`/`j`/`i`) plus a single affine (or, for multi-pose data, one affine per pose), instead of axis-aligned physical `z`/`y`/`x` coordinates. World coordinates are derived lazily from that affine via a custom xarray index (`VoxelToWorldIndex`), never stored directly. This lets ConfUSIus represent oblique/rotated/sheared acquisitions and registration outputs exactly, without resampling onto an axis-aligned grid — previously either lossy or impossible — and extends naturally to multi-pose probe sweeps where each pose sits at a genuinely different physical position.

## Breaking changes

- **Canonical dims**: `(...extra, time, pose, z, y, x)` → `(...extra, time, pose, k, j, i)`. `z`/`y`/`x` are derived, not stored. Every loader, registration function, plotting path, napari layer, and I/O round trip was migrated; `validate_fusi`/`ensure_fusi` enforce this unconditionally, no fallback to a plain shape.
- Renamed "physical" → "world" throughout the API (`attrs["affines"]` keys, function/parameter names, docs); `orientation_matrix` → `direction_matrix`; `affine_to` → `get_relative_affine`.
- `cf.io.load`'s `.zarr` branch and the napari Zarr reader now reject stores not written by `cf.io.save`, instead of silently loading them as non-canonical.
- `extract_with_mask`/`extract_with_labels`/`validate_atlas` now require canonical VoxelData input and check alignment via the full affine, not just voxel-index ranges.
- `pose` is its own independently indexed coordinate now (not jointly owned by `VoxelToWorldIndex`) — fixes a real `AlignmentError` bug after `.sel(pose=idx)`/`.isel(pose=idx)`. One consequence: combined `.sel(pose=0, z=..., y=..., x=...)` needs two calls now; `.isel()` is unaffected.
- Multi-pose `time` is now `(time, pose)`-shaped, holding each pose's own real timestamp directly, replacing the old `pose_time` sidecar.
- `consolidate_poses` dropped `affines_key`; always reads from primary geometry now.

## Enhancements

- Pose-dependent voxel-to-world geometry (`(npose, 4, 4)` affine stacks) throughout the model, plus new [`stack_poses`][confusius.multipose.stack_poses] to assemble independently loaded single-pose grids into one pose-dependent DataArray.
- New `.fusi.affine` accessor (`apply`, `reindex_voxels`, `reindex_voxels_like`) for moving VoxelData between voxel/world/named-affine spaces.
- NIfTI/Zarr I/O round-trip oblique geometry exactly.
- `plot_volume`/napari can slice by native voxel plane or world plane, and facet over `pose`.

## Fixes

Several correctness bugs surfaced along the way: displacement-field/atlas composition mixing voxel-index values into world-mm math, `SearchLight` radius in the wrong units, GLM registration initialization, NIfTI qform/sform round-tripping, a pose-stacked affine misindexing bug, and a `plot_volume` axis mislabeled "(mm)" for `pose`.

See `docs/changelog.md`'s `0.7.0.dev0` section for the full, itemized list.


FelipeCybis (approved):
Should we merge this and continue easier and lighter development on main?

sdiebolt:
I think we should indeed, otherwise the PR will keep on growing as we find stuff to add to the model 😅

That's the problem with a data model, there's always stuff that seem related haha

PR #279, opened 2026-07-09
feat(datasets): add citation info to dataset fetchers

FelipeCybis:
## Summary

Adds citation information to all five dataset/template fetchers so users know what to cite when they use the data.

- Each fetcher now prints `Fetched data from <citation>.` after a successful fetch, controlled by a new `show_citation_msg: bool = True` parameter. Pass `show_citation_msg=False` to silence it (e.g. in scripts and notebooks).
- The two template fetchers (`fetch_template_huang_2025`, `fetch_template_pepe_mariani_2026`) additionally stamp the citation onto the returned DataArray as `da.attrs["citation"]`, so it survives as metadata regardless of the print flag.
- Citation strings are stored as a module-level `_CITATION` constant in each dataset module.

## Fetchers affected

`fetch_cybis_pereira_2026`, `fetch_landemard_2026`, `fetch_nunez_elizalde_2022`, `fetch_template_huang_2025`, `fetch_template_pepe_mariani_2026`.

## Tests

- Added a citation test per fetcher: BIDS fetchers assert the message prints by default and is suppressed under `show_citation_msg=False`; template fetchers additionally assert `da.attrs["citation"]` is set.
- Fixed the template fetchers' `mock_load` fixture, whose bare `object()` sentinel could not hold the newly-written `attrs["citation"]` — it now uses `SimpleNamespace(attrs={})`.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,12 @@ Current development version for the next ConfUSIus release.
 
 ### :sparkles: Enhancements
 
+- Dataset fetchers now print the citation to use for the fetched data and accept a
+  `show_citation_msg` argument to silence it. The template fetchers
+  [`fetch_template_huang_2025`][confusius.datasets.fetch_template_huang_2025] and
+  [`fetch_template_pepe_mariani_2026`][confusius.datasets.fetch_template_pepe_mariani_2026]
+  also expose the citation on the returned DataArray as `da.attrs["citation"]`
+  ([#279](https://github.com/confusius-tools/confusius/pull/279)).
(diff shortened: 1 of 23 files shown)


FelipeCybis:
@sdiebolt  Good to be reviewed!

FelipeCybis:
@sdiebolt there was just a cache thing that prevented one example to have the new outputs. I think when the next CI passes everything will be good. The word wrapping of the rich outputs (and stderr outputs) are being handled in #285

sdiebolt (approved):

PR #284, opened 2026-07-09
fix(io): make Zarr saving handle non-serializable attrs

### LINKED ISSUE #283: Saving data to zarr with non-serializable attributes (e.g. affines) fails
### ConfUSIus version

v0.5.0

### Python version

3.14

### Operating system

WSL

### Installation

uv

### Description

Saving data to `zarr` with affines (or any non-serializable attributes) fails.

### Steps to reproduce

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
# ]
# ///
#
# for issues.

import confusius as cf

bids_root = cf.datasets.fetch_nunez_elizalde_2022(
    subjects="CR022",
    sessions="20201011",
    tasks="spontaneous",
    acqs="slice03",
)
pwd_path = (
    Path(bids_root)
    / "sub-CR022"
    / "ses-20201011"
    / "fusi"
    / "sub-CR022_ses-20201011_task-spontaneous_acq-slice03_pwd.nii.gz"
)
data = cf.load(pwd_path)
data.fusi.save("to_zarr.zarr")

```
We get a 
```
TypeError: Invalid attribute in Dataset.attrs.
```

### Additional output

_No response_

FelipeCybis:
## Summary

Closes #283.

Saving a fUSI DataArray to Zarr failed whenever its attrs held values Zarr can't JSON-encode — most commonly `attrs["affines"]` (a dict of numpy arrays) or matplotlib `cmap`/`norm` objects on atlas-derived data. `data.fusi.save("out.zarr")` raised `TypeError: Invalid attribute in Dataset.attrs.`

Root cause: Zarr stores attrs as JSON. It already tolerates *top-level* numpy arrays and scalars, but not numpy arrays nested inside a dict (exactly how affines are stored), nor arbitrary Python objects.

## Changes

`save()` now sanitizes a shallow copy of the DataArray's attrs before `to_zarr` (the caller's object is left untouched):

- Nested numpy arrays and scalars are recursively converted to native Python (JSON-storable) — so affines survive as lists.
- Any value still not JSON-encodable is dropped with a warning. Atlas `cmap`/`norm` fall in this bucket and are already rebuilt from `rgb_lookup` on load.

Unlike the NIfTI sidecar path, this deliberately skips the NIfTI-specific machinery: no BIDS field renaming, no `dim4`/`dim5` or timing metadata.

On load, `attrs["affines"]` dict values are restored to numpy arrays, so a Zarr round-trip matches the NIfTI and SCAN loaders' contract. This is a no-op for those loaders (their affines are already numpy arrays).

## Testing

- New round-trip tests: affines reload as numpy arrays with correct values; non-serializable attrs are dropped with a warning and the rest survive; the input DataArray is not mutated.
- Reworked the two zarr dispatch tests into real write-and-reload checks (the old ones asserted `to_zarr` was called on the raw object, which no longer holds now that we copy + sanitize first).
- Verified end-to-end against the issue's real `pwd.nii.gz -> .zarr` round-trip; confirmed NIfTI load is unchanged (affines stay `ndarray`/`float64`).
- All `tests/unit/test_io` pass; ruff, ty, codespell, and numpydoc-validation clean.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -36,6 +36,11 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- Saving to Zarr (via [`save`][confusius.io.save] or `DataArray.fusi.save`) now works
+  for data carrying affines or other numpy-valued attributes: nested numpy arrays are
+  stored as lists and non-serializable attrs (e.g. matplotlib colormaps) are dropped
+  with a warning, matching the NIfTI sidecar behaviour
+  ([#283](https://github.com/confusius-tools/confusius/issues/283)).
 - B-spline control-point DataArrays returned by
   [`register_volume`][confusius.registration.register_volume] no longer have their
   per-axis grid geometry (spacing, origin, domain) swapped between axes on anisotropic
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -40,7 +40,7 @@ Current development version for the next ConfUSIus release.
   for data carrying affines or other numpy-valued attributes: nested numpy arrays are
   stored as lists and non-serializable attrs (e.g. matplotlib colormaps) are dropped
   with a warning, matching the NIfTI sidecar behaviour
-  ([#283](https://github.com/confusius-tools/confusius/issues/283)).
+  ([#284](https://github.com/confusius-tools/confusius/pull/284)).
 - B-spline control-point DataArrays returned by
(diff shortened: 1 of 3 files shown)


FelipeCybis:
@sdiebolt ready to review!

sdiebolt (approved):
LGTM! (With a small patch to get back to 100% coverage)

PR #285, opened 2026-07-09
feat(docs): run individual gallery examples and wrap long output

### LINKED ISSUE #280: Gallery tools should be able to run individual examples
Today our gallery build tool discovers all python file under the examples folder and build the examples from them indiscriminately. We should have more freedom to build individual examples.

I am thinking about adding a more CLI friendly arguments to the build_gallery tools. If no path is given, discover all example files in a default folder. But we should be possible to pass paths with `uv run tools/build_gallery.py docs/examples/01_io/01_confusius_xarray_101.py docs/examples/02_registration/01_register_volume_same_subject.py`.

FelipeCybis:
Closes #280.

## Summary

Adds arguments to the gallery builder so you can iterate on individual examples without paying the full-gallery build time, and fixes two cases where long example output showed a horizontal scrollbar instead of wrapping.

## Run individual examples

`tools/build_gallery.py` now accepts optional example paths. With no arguments it runs every example as before; with paths it runs only those:

```
uv run python tools/build_gallery.py docs/examples/01_io/01_confusius_xarray_101.py docs/examples/03_decomposition/01_pca_single_recording.py
```

The whole gallery is still rendered either way, so `index.md` keeps listing everything. For each example not named on the command line:

1. if it is already in the cache, it is restored from cache (full page with outputs);
2. otherwise it is rendered from source **without running** — code and markdown cells only, no outputs — and is deliberately **not** cached, so a later full or targeted build still executes it.

Unknown paths fail fast with a message listing them.

## Wrap long output instead of scrolling

- Rich-rendered text — e.g. the dataset citation banner the fetchers print — embeds as a monospace block with an inline `white-space:pre; overflow-x:auto`. A CSS rule scoped to rich's inline `font-family:Menlo…` wraps it and breaks long unbreakable tokens such as DOI URLs. xarray HTML reprs (which should keep scrolling) are untouched.
- Plain-text output — warnings, `text/plain` reprs, tracebacks — is now wrapped in a `<div class="gallery-output" markdown>` container (Zensical's `md_in_html` keeps the highlighted code-block styling), and CSS wraps its lines. Input `python` code cells and xarray text reprs elsewhere in the docs still scroll.

## Testing

- New/updated tests: builds only selected examples while stubbing the rest (asserts the selected example has output, the rest have code but no `gallery-output`, aren't cached, and the index lists both); unknown-path validation; render test asserting output is wrapped while input code stays a bare highlighted block.
- Verified end-to-end against the real examples: running one example executes it, reuses a cached example, stubs the rest, and lists all seven in the index. Confirmed the citation-banner CSS selector against the pr-279 preview page (matches the one rich-text block, leaves the three xarray reprs alone).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -74,7 +74,7 @@ Current development version for the next ConfUSIus release.
   ([#235](https://github.com/confusius-tools/confusius/pull/235)).
 - Long output in gallery examples — warnings, text reprs, tracebacks, and rich-rendered
   text such as the dataset citation banner — now wraps instead of showing a horizontal
-  scrollbar ([#284](https://github.com/confusius-tools/confusius/pull/284)).
+  scrollbar ([#285](https://github.com/confusius-tools/confusius/pull/285)).
 
 ### :wrench: Maintenance
 
@@ -84,7 +84,7 @@ Current development version for the next ConfUSIus release.
   (`uv run python tools/build_gallery.py docs/examples/01_io/01_confusius_xarray_101.py`),
   running only those; the rest of the gallery is still rendered, taken from cache if
   present or built without outputs
-  ([#284](https://github.com/confusius-tools/confusius/pull/284)).
+  ([#285](https://github.com/confusius-tools/confusius/pull/285)).
 
 ## 0.5.0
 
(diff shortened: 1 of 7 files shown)


FelipeCybis:
@sdiebolt this is ready to review!

sdiebolt (approved):
LGTM!

PR #286, opened 2026-07-10
maint: replace broad dependency extras

### LINKED ISSUE #168: Narrow heavy optional dependencies in core install
Core install currently pulls in very broad transitive dependency stacks via:

- `napari[all]>=0.7.0`
- `xarray[complete]>=2025.1.0`

This can make installation fragile across Python/platform combinations. In my case, project locking selected an older `numba` path (`0.62.1`), which forced `llvmlite==0.45.1` and failed to build on Python 3.14, even though a single-environment install succeeded with newer `numba/llvmlite`.

sdiebolt:
## Summary
- replace `napari[all]` with explicit napari/Qt/network deps we actually want
- replace `xarray[complete]` with explicit acceleration deps we actually want
- drop unrelated transitive baggage from the core install

Closes #168


FelipeCybis (commented):
I have a bit of mixed feelings about this. For `xarray[complete]` there is actually a bunch of dependencies that we really do not use, but for `napari[all]` isn't this going to add extra burden on us for controlling everything? Specially if breaking changes happen upstream by adding or changing dependencies on the groups that we use?

sdiebolt:
After sleeping on it I agree with you. Even for Xarray, I'd prefer having their own specific dependencies (eg. Numbagg) handled by them. I'll modify this PR or close it if I can't find something.

sdiebolt:
@FelipeCybis Should be better now, what do you think?

FelipeCybis:
> @FelipeCybis Should be better now, what do you think?

Seems good to me now! Isn't there something we should keep doing with `numba`, though?

sdiebolt:
I thought so originally... It was for a macOS dependency problem: napari defines
```
    "numba>=0.58.0,<=0.62.1 ; platform_system == 'Darwin' and platform_machine == 'x86_64'",
    "numba>=0.58.0 ; platform_system != 'Darwin' or platform_machine != 'x86_64'",
```
and it was causing dependency errors on macOS runners in CI because numba 0.62.1 doesn't support Python 3.14... But somehow macOS seems to download numba 0.66 now, so I guess the macOS runners have all migrated to ARM? Which would make sense I guess, there must not be many Intel macs out there anymore. I'd leave it as is for now and we'll see if we encounter problems with macOS users?

FelipeCybis (approved):

FelipeCybis:
> I thought so originally... It was for a macOS dependency problem: napari defines
> 
> ```
>     "numba>=0.58.0,<=0.62.1 ; platform_system == 'Darwin' and platform_machine == 'x86_64'",
>     "numba>=0.58.0 ; platform_system != 'Darwin' or platform_machine != 'x86_64'",
> ```
> 
> and it was causing dependency errors on macOS runners in CI because numba 0.62.1 doesn't support Python 3.14... But somehow macOS seems to download numba 0.66 now, so I guess the macOS runners have all migrated to ARM? Which would make sense I guess, there must not be many Intel macs out there anymore. I'd leave it as is for now and we'll see if we encounter problems with macOS users?

Fine by me. A bit offended looking at my _vintage_ 2017 intel mac under my night stand.

PR #290, opened 2026-07-10
fix: avoid py3.14 macOS Intel numba sync failure

sdiebolt:
Python 3.14 on macOS Intel was pulling into the xarray[accel]/numba/llvmlite path, but napari's macOS Intel numba cap kept resolution on versions that fail to build there.

This keeps xarray[accel] everywhere else and falls back to plain xarray only for Python 3.14 on macOS Intel.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,9 +6,15 @@ icon: lucide/history
 
 # Changelog
 
-## 0.6.0.dev0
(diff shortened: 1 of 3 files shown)


sdiebolt:
@FelipeCybis Quick hotfix for 0.5.2, ready to review

FelipeCybis (approved):

PR #294, opened 2026-07-10
fix: support BIDS physio import in napari

### LINKED ISSUE #293: Support BIDS physio import in napari signal plotter
The napari signal importer currently expects a `time` column plus signal columns. BIDS physio files are usually `*.tsv.gz` without headers, with `Columns`, `SamplingFrequency`, and optional `StartTime` stored in the JSON sidecar.

Suggestion: detect the matching sidecar for physio TSV files and synthesize the `time` column from the sidecar metadata when present.

sdiebolt:
Closes #293.

Adds BIDS physio `*.tsv.gz` import support to the napari signal store by reading `Columns`, `SamplingFrequency`, and optional `StartTime` from the JSON sidecar.

Renamed `read_`/`write_` functions in BIDS `events.py` to `load_`/`save_` to match other I/O functions in ConfUSIus.

### DIFF
--- a/tests/unit/test_napari/test_signal_store.py
+++ b/tests/unit/test_napari/test_signal_store.py
@@ -52,19 +52,6 @@ def test_import_bids_physio_tsv_gz_uses_sidecar_columns_and_time(
     npt.assert_array_equal(imported[1].y, np.array([10.0, 11.0, 12.0]))
 
 
-def test_import_bids_physio_rejects_column_count_mismatch(signals_store, tmp_path):
-    path = tmp_path / "sub-01_task-rest_recording-cardiac_physio.tsv.gz"
-    with gzip.open(path, "wt") as f:
-        f.write("1\t10\n2\t11\n")
-
-    path.with_suffix("").with_suffix(".json").write_text(
-        json.dumps({"Columns": ["pulse"], "SamplingFrequency": 10.0})
-    )
-
-    with pytest.raises(ValueError, match="Columns' length does not match TSV width"):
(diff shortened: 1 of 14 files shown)


sdiebolt:
@FelipeCybis This should be ready to review. I've introduced a breaking change by renaming `read_`/`write_` functions into `load_`/`save_` to match other I/O functions in ConfUSIus, but I'd be happy to have your opinion on this!

tests/unit/test_napari/test_signal_store.py
```diff
@@ -23,6 +26,59 @@ def test_import_tsv_preserves_duplicate_time_values(signals_store, signals_tsv):
...
+        f.write("1\t10\n2\t11\n")
+
+    path.with_suffix("").with_suffix(".json").write_text(
+        json.dumps({"Columns": ["pulse"], "SamplingFrequency": 10.0})
+    )
+
+    with pytest.raises(ValueError, match="Columns' length does not match TSV width"):
+        signals_store.import_file(path)
+
+
```
FelipeCybis: This test is also added in the bids module.

```suggestion
```

FelipeCybis (approved):
I haven't been able to play around on Napari (also I don't have such type of data 😄), but the logic sounds sane to me. There is just a repeated test that I would suggest removing.

sdiebolt:
Indeed, that might be because I had originally implemented the feature in the napari module only, then thought it best to have it in the BIDS module, thanks for catching it! 😅

PR #296, opened 2026-07-10
fix(plotting): accept scalar slice dimension

### LINKED ISSUE #295: Plotting functions should be able to slice through a scalar dimension

A `slice_mode` that slice through a scalar dimension should be accepted and work no difference then a degenerated dimension. For example:

```python
cf.plotting.plot_contours(atlas.annotation.sel(z=6))
```

FelipeCybis:
- Closes #295.

## Description

A scalar slice_mode coordinate is now promoted back to a size-1 dimension via expand_dims, so sel(z=6) plots identically to sel(z=[6]). This is applied at the two normalization sites: _prepare_slice_inputs (covers plot_volume/plot_image and plot_comparison) and the mask path in add_contours (covers plot_contours), via a shared _is_scalar_coord helper.

Invalid slice modes are unaffected: slice_mode="t" on data with no t coordinate still raises, because promotion only triggers when the name exists as a 0-dimensional coordinate.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -10,6 +10,12 @@ icon: lucide/history
 
 Current development version for the next ConfUSIus release.
 
+### :bug: Fixes
+
+- Plotting functions now accept a slice dimension reduced to a scalar coordinate by a
+  single-index selection, so `plot_contours(atlas.annotation.sel(z=6))` works like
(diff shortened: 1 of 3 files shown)


FelipeCybis:

sdiebolt (approved):
LGTM!

sdiebolt (approved):
LGTM!

PR #299, opened 2026-07-11
fix: clarify SCAN v2 load error

### LINKED ISSUE #297: Raise useful error when trying to open SCAN v2 files
Without documentation from Iconeus or reverse engineering, we can't support the new SCAN v2 format released recently by Iconeus. In the meantime, it would be helpful to raise a useful error when a user opens a `.scan` file that isn't recognized as an HDF5, hinting at the fact that it may be a SCAN v2 file that can be converted to NIfTI using Iconeus tools

sdiebolt:
Closes #297.

Adds a clear error for non-HDF5 \.scan files and points users to Iconeus conversion to NIfTI.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,9 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- Opening a `.scan` file that is not the legacy HDF5-based Iconeus format now raises a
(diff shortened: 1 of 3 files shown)


sdiebolt:
@FelipeCybis Good to review!

FelipeCybis (approved):
LGTM!

PR #301, opened 2026-07-12
fix(registration): relabel motion params by dim

### LINKED ISSUE #300: create_motion_dataframe mislabels motion parameters when spatial dims are not x/y/z
`create_motion_dataframe()` assumes affine parameter order is always `(x, y, z)`, but the transform components follow the DataArray spatial dim order. For arrays with spatial dims like `(z, y, x)`, both translation (`trans_*`) and rotation (`rot_*`) columns are mislabeled. Example: for a `(z, y, x)` single-slice recording, in-plane x motion appears in `trans_z`, not `trans_x`.

sdiebolt:
Fixes #300.

Maps `rot_*` and `trans_*` columns onto the DataArray's named spatial axes instead of raw transform-component order.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,7 +12,12 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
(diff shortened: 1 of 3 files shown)


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (approved):
LGTM!

PR #216, opened 2026-06-26
feat(napari): add registration panel

sdiebolt:
This PR adds a new napari Registration tab for between-scan and within-scan workflows. It supports live previews and progress reporting, cooperative aborts with partial results, transform save/load/apply flows (including manual napari transforms and B-spline control-point grids via Zarr), and the registration parameter controls needed to run the existing ConfUSIus registration backends from the GUI.

Closes #210.
Closes #213.
Closes #171.

PR #302, opened 2026-07-12
docs(docs): add motion correction example

### LINKED ISSUE #98: [New example]: Registration
- [x] #139 
- [x] Registration to the Pepe Mariani 2026 template for Allen alignment (done in #243 and #242).
- [x] #282
- [x] #281

### LINKED ISSUE #282: Add volumewise registration example
This example can be done with data from `fetch_cybis_pereira_2026`.

Things that would be nice to show:
- Diagnostics from the `confusius.registration.motion.create_motion_dataframe`
- GIF for before/after registration
- Line plot from a given voxel before/after registration.

sdiebolt:
Adds a gallery example for volumewise motion correction using the same dataset and registration settings as the GUI registration demo.

Also folds in the motion-diagnostics cleanup that the example needed:
- add `confusius.plotting.plot_motion_diagnostics()` for plotting motion summaries
- make `create_motion_dataframe()` label motion columns by named axes
- keep the full affine-dimensional motion summary in the output table (`rotation` + translations for 2D affines, full rotation/translation axes for 3D affines even when one spatial axis is singleton)
- require real affine arrays in motion-diagnostics helpers instead of `None` placeholders
- move Euler-angle extraction to `confusius.registration.affines` and cover the new behavior with unit tests

Closes #282
Closes #98

### DIFF
--- a/src/confusius/registration/affines.py
+++ b/src/confusius/registration/affines.py
@@ -125,6 +125,46 @@ def compose_affine(
     return A
 
 
+def get_euler_xyz_from_rotation_matrix(
+    R: npt.NDArray[np.floating],
+) -> tuple[float, float, float]:
+    """Get XYZ Euler angles from a 3D rotation matrix.
+
+    Parameters
+    ----------
+    R : (3, 3) numpy.ndarray
+        Rotation matrix.
+
+    Returns
+    -------
+    rot_x : float
+        Rotation angle around the x axis in radians.
+    rot_y : float
+        Rotation angle around the y axis in radians.
+    rot_z : float
+        Rotation angle around the z axis in radians.
+    """
+    if R.shape != (3, 3):
+        raise ValueError(f"Expected a (3, 3) rotation matrix, got {R.shape}.")
+
+    # XYZ convention: R = Rz @ Ry @ Rx
+    # R[2,0] = -sin(rot_y)
+    sy = -R[2, 0]
+    sy = max(-1.0, min(1.0, sy))
+    rot_y = math.asin(sy)
+
+    cos_y = math.cos(rot_y)
+    if abs(cos_y) > 1e-6:
+        rot_x = math.atan2(R[2, 1], R[2, 2])
+        rot_z = math.atan2(R[1, 0], R[0, 0])
+    else:
+        # Gimbal lock: set rot_x = 0 and compute rot_z.
+        rot_x = 0.0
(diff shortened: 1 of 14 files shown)


sdiebolt:
@FelipeCybis Ready to review! This should be merged before #216, I'll merge it in then.

src/confusius/registration/motion.py
```diff
@@ -6,99 +6,124 @@
...
-            # Rotation angle from the 2D rotation matrix.
             rotation = math.atan2(R[1, 0], R[0, 0])
             params_list.append([rotation, float(T[0]), float(T[1])])
         else:
-            # Decompose the 3D rotation matrix into Euler angles (XYZ convention).
-            rot_x, rot_y, rot_z = _rotation_matrix_to_euler_xyz(R)
+            rot_x, rot_y, rot_z = get_euler_xyz_from_rotation_matrix(R)
             params_list.append(
                 [rot_x, rot_y, rot_z, float(T[0]), float(T[1]), float(T[2])]
             )
```
FelipeCybis: Should we use `rot_0, rot_1, rot_2` (or just `rot = get_euler_xyz_from_rotation_matrix(R)` and append `rot[0], rot[1], rot[2]` to avoid confusion with the callers of `extraction_motion_parameters`, that will usually pass affines with row-transform order of `zyx`?

src/confusius/registration/affines.py
```diff
@@ -125,6 +125,46 @@ def compose_affine(
     return A
 
 
+def get_euler_xyz_from_rotation_matrix(
```
FelipeCybis: Should this really be public? I am not sure about it.

FelipeCybis (commented):
A part of some commits I have made, here are my comments

src/confusius/registration/affines.py
sdiebolt: Thought it nice to have, but probably a bit too niche. And users can always use the transforms3d package. I'll put it private.

src/confusius/registration/motion.py
sdiebolt: The thing is, `sxyz`/`rxyz`/`szyx`/`rzyx` are conventional names for rotation composition ordering. But I agree that we should probably rename the return parameters. Will fix.

FelipeCybis (approved):
LGTM

PR #304, opened 2026-07-12
fix(io): handle bad volume timing sidecars

### LINKED ISSUE #303: NIfTI load fails hard when sidecar timing length mismatches data
When a NIfTI sidecar JSON contains timing metadata that does not match the actual data shape, ConfUSIus raises an `xarray.core.coordinates.CoordinateValidationError` during load and refuses to open the dataset.

sdiebolt:
## Summary
- ignore malformed `VolumeTiming` when its length disagrees with the data
- fall back to `pixdim[4]`, then to frame indices with a dedicated warning
- add regression tests for header and frame-index fallbacks

Closes #303

### DIFF
--- a/src/confusius/io/nifti.py
+++ b/src/confusius/io/nifti.py
@@ -429,7 +429,7 @@ def _create_spatial_coords_from_nifti(
 
 def _validate_volume_timing_length(
     volume_timing: npt.NDArray[np.floating], *, n_time: int
-) -> npt.NDArray[np.floating] | None:
+) -> tuple[npt.NDArray[np.floating] | None, bool]:
     """Validate sidecar `VolumeTiming` against the NIfTI time-axis length.
 
     Parameters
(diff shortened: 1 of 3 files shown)


sdiebolt:
@FelipeCybis Ready to review!

src/confusius/io/nifti.py
```diff
@@ -490,16 +526,27 @@ def _create_temporal_coords_from_nifti(
...
-                raise_on_unknown=True,
-            )
-    elif "repetition_time" in attrs:
+        raw_volume_timing = np.asarray(attrs.pop("volume_timing"), dtype=np.float64)
+        volume_timing_length_mismatch = (
+            raw_volume_timing.ndim == 1
+            and raw_volume_timing.size > 0
+            and raw_volume_timing.size != n_time
+        )
+        volume_timing = _validate_volume_timing_length(raw_volume_timing, n_time=n_time)
```
FelipeCybis: Can't we get the length mismatch from the validation function? Maybe returning `volume_timing` and `mismatch: bool`? It seems that we are doing the same check twice

sdiebolt: We can definitely absorb that into the validation function!

FelipeCybis (approved):
LGTM

PR #307, opened 2026-07-13
docs(gallery): style pandas DataFrame reprs like Jupyter notebooks

### LINKED ISSUE #306: Improve pandas rendering in example gallery
Currently

image

But it should be something like

image

If I understood correctly, we are just not getting the `__repr_html__` for pandas outputs.

FelipeCybis:
Fixes #306.

## Summary

pandas DataFrame outputs in the example gallery rendered as an ugly fully-bordered grid. The renderer already prefers `text/html`, so it *was* emitting pandas' `__repr_html__` — but pandas ships `<table border="1" class="dataframe">` with only a tiny scoped `<style>`, and the docs theme styles unclassed tables only (`.md-typeset table:not([class])`). With no CSS for the classed `.dataframe` table, the browser fell back to the `border="1"` grid.

This adds theme-aware CSS for `.gallery-rich-output .dataframe` so the table renders in the clean Jupyter notebook look — centered, with alternating row stripes reusing the xarray coordinate repr shades and a mouse-hover row highlight — in both light and dark schemes. CSS-only; no Python change.

image

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -19,6 +19,12 @@ Current development version for the next ConfUSIus release.
   single-index selection, so `plot_contours(atlas.annotation.sel(z=6))` works like
   `sel(z=[6])` ([#296](https://github.com/confusius-tools/confusius/pull/296)).
 
+### :wrench: Maintenance
+
+- pandas DataFrame outputs in the example gallery now render with clean, theme-aware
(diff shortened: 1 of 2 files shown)


FelipeCybis:

sdiebolt (approved):
LGTM!

PR #313, opened 2026-07-14
refactor(iq): simplify axial velocity estimation

### LINKED ISSUE #314: Let axial velocity spatial_kernel accept per-axis sizes and default to 3
Axial velocity currently treats `spatial_kernel` as one scalar applied equally to `(z, y, x)`, in voxel units.

We should:
- allow `spatial_kernel` to be either an int or a `(z, y, x)` list/tuple;
- keep scalar input as shorthand for the same size on all axes;
- change the default from `1` to `3`.

Reason: Kasai velocity is noisy enough that spatial median filtering is usually needed, and anisotropic data should be able to avoid over-filtering along thin axes like `z`.

sdiebolt:
Always uses the standard Kasai estimator, removes `estimation_method`, and lets axial velocity `spatial_kernel` default to `3` or be passed per axis as `(z, y, x)`.

Closes #314.

### DIFF
--- a/src/confusius/iq/process.py
+++ b/src/confusius/iq/process.py
@@ -802,7 +802,6 @@ def compute_axial_velocity_volume(
     spatial_kernel: int = 1,
     transmit_frequency: float = 15.625e6,
     beamforming_sound_velocity: float = 1540,
-    estimation_method: Literal["average_angle", "angle_average"] = "average_angle",
 ) -> npt.NDArray:
     """Compute axial velocity volumes from beamformed IQ data.
 
@@ -858,13 +857,6 @@ def compute_axial_velocity_volume(
         Ultrasound transmit frequency in hertz.
     beamforming_sound_velocity : float, default: 1540
         Speed of sound assumed during beamforming, in meters per second.
-    estimation_method : {"average_angle", "angle_average"}, default: "average_angle"
-        Method for computing the velocity estimate.
-
-        - `"average_angle"`: Compute the angle of the autocorrelation, then average
-          (i.e., average of angles).
-        - `"angle_average"`: Average the autocorrelation, then compute the angle
-          (i.e., angle of average).
 
     Returns
     -------
@@ -887,7 +879,6 @@ def compute_axial_velocity_volume(
         fs: float,
         transmit_frequency: float,
         beamforming_sound_velocity: float,
-        estimation_method: Literal["average_angle", "angle_average"],
         **_,
     ) -> npt.NDArray:
         """Compute the Kasai estimator from an array of IQ data.
@@ -910,8 +901,6 @@ def compute_axial_velocity_volume(
             Ultrasound transmit frequency in hertz.
         beamforming_sound_velocity : float
             Speed of sound in the imaged medium, in meters per second.
-        estimation_method : {"average_angle", "angle_average"}
-            Method for computing the velocity estimate.
         **_
             Additional unused keyword arguments (absorbed and ignored).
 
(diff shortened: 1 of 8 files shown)


sdiebolt:
@FelipeCybis ready to review!

src/confusius/iq/process.py
FelipeCybis, line 974: Actually, I think we are not correctly scaling for lag > 1 here, are we? The denominator should be scaled by `* lag`.

I was not sure so I asked for synthetic iq data to Claude and made this script:
```python
from typing import cast

import numpy as np
import numpy.typing as npt

C = 1540.0  # speed of sound, m/s
F0 = 15e6  # ultrasound frequency, Hz
FRAMERATE = 500.0  # volume rate, Hz
V_TRUE = 2e-3  # true axial speed, m/s (toward the probe)


def process_block_func(
    block: npt.NDArray,
    lag: int,
    spatial_kernel: int | tuple[int, int, int] | list[int] = 1,
    fs: float = FRAMERATE,
    transmit_frequency: float = F0,
    beamforming_sound_velocity: float = C,
    **_,
) -> npt.NDArray:
    """Compute the Kasai estimator from an array of IQ data.
    ----------
    block : (time, z, y, x) numpy.ndarray
        Complex IQ data.
    spatial_kernel : int or tuple[int, int, int] or list[int]
        Size of the median filter kernel applied spatially to denoise. A scalar
        uses the same size on all spatial axes; a length-3 sequence specifies
        `(z, y, x)` sizes directly.
    lag : int
        Temporal lag in volumes for autocorrelation computation.
    fs : float
        Volume sampling frequency in hertz.
    transmit_frequency : float
        Ultrasound transmit frequency in hertz.
    beamforming_sound_velocity : float
        Speed of sound in the imaged medium, in meters per second.
    **_
        Additional unused keyword arguments (absorbed and ignored).

    Returns
    -------
    (z, y, x) numpy.ndarray
        Axial velocity volume in meters per second.
    """
    block_rolled_conjugate = np.roll(block, lag, axis=0).conj()
    block_rolled_conjugate[:lag, ...] = 0
    autocorrelation = cast("npt.NDArray", block * block_rolled_conjugate)[lag:]

    average_autocorrelation_phase = np.angle(autocorrelation.mean(0))

    return (
        average_autocorrelation_phase
        * fs
        * beamforming_sound_velocity
        / (4 * np.pi * transmit_frequency)
    )


def synthetic_iq(v, n_frames=200, shape=(6, 5, 4), noise=0.05, seed=0):
    """IQ volume in which every voxel moves toward the probe at `v` m/s.

    Round-trip baseband phase of a scatterer at depth d, approaching at speed v
    so that d(n) = d0 - v * n / framerate:

        dphi   = 4 pi f0 v / (c * framerate)          # rad per frame, > 0 toward probe
        z(n)   = exp(1j * dphi * n) + noise

    """
    rng = np.random.default_rng(seed)
    dphi = 4 * np.pi * F0 * v / (C * FRAMERATE)

    z = np.broadcast_to(np.exp(1j * dphi * np.arange(n_frames)), (*shape, n_frames)).T
    return z + noise * (rng.normal(size=z.shape) + 1j * rng.normal(size=z.shape))


data = synthetic_iq(V_TRUE)

for lag in (1, 2, 3, 4):
    v_bug = process_block_func(data, lag=lag).mean() * 1e3
    print("Axial velocity:", v_bug, "mm/s")
```
```
Axial velocity: 2.00003895573522 mm/s
Axial velocity: 3.9999201733660947 mm/s
Axial velocity: 5.999489422567463 mm/s
Axial velocity: 7.999693517093211 mm/s
```

sdiebolt: Indeed, that's totally a bug, thanks for catching it!

FelipeCybis (approved):
LGTM

PR #317, opened 2026-07-15
feat(io): add experimental Iconeus SCAN v2 loader

### LINKED ISSUE #316: feat(io): add Iconeus SCAN v2 (binary `.scan`) loader
## Context

`load_scan` currently only handles the HDF5-based Iconeus SCAN format and explicitly rejects the newer binary SCAN v2 files (see #297).

Some ConfUSIus users have reported needing to open SCAN v2 files. We were kindly provided an example SCAN v2 file and have been able to map out a large part of the header and load the payload correctly.

## What we found

The file is a flat binary: a variable-size header followed by a little-endian `float64` payload. Header size and payload size are both stored in the header, and the payload byte count matches the product of the dimension fields exactly:

```
sizeX * sizeY * sizeZ * nTime * npose * nblockRepeat * 8 = payload bytes
```

Decoded fields:

| Offset | Type | Field | Value |
|---|---|---|---|
| `0x00` | char[4] | magic | `scan` |
| `0x04` | u64 | version | `1` |
| `0x0C` | u64 | fixed header bytes |  |
| `0x14` | u64 | metadata block bytes |  |
| `0x1C` | u32 | unknown |  |
| `0x20` | u64 | total header bytes |  |
| `0x28` | u64 | payload bytes |  |
| `0x5C` | u64 | `sizeX` (lateral) | `128` |
| `0x64` | u64 | `sizeY` (elevation) | `1` |
| `0x6C` | u64 | `sizeZ` (depth) | `92` |
| `0x74` | u64 | `nTime` | `38` |
| `0x7C` | u64 | `npose` (?) | `1` |
| `0x84` | u64 | `nblockRepeat` (?) | `1` |
| `0x94` | f64 | dt / sampling period (s) | `0.4` |
| `0x9C` | u32 | SVD clutter low cutoff | `60` |
| `0xA0` | u32 | power Doppler integration window | `200` |
| `0xA4` | f64 | x voxel size (m) | `0.00011` |
| `0xAC` | f64 | y voxel size (m) | `0.0004` |
| `0xB4` | f64 | z voxel size (m) | `0.00009856` |
| `0xD4` | f64[nTime] | time coordinates (s) | `0.4, 0.8, ...` |
| `0x204` | u32[nTime] | frame indices | `0 … nTime - 1` |
| `0x29C` | f64[8] | 6DOF probe pose (tx,ty,tz,rx,ry,rz + flag + pad) | `0.0007, 0.0026, 0, -0, 0, π/2, 1, 0` |
| `0x2DC` | f64 | probe center frequency (MHz) | `15.625` |
| `0x2E4` | f64 | probe pitch (mm) | `0.11` |
| `0x2EC` | f64 | scan depth (cm) | `1.5` |
| `0x2F4` | f64 | elevation focus (mm) | `8.0` |
| `0x30E` | str | probe name | `IcoPrime` |
| `0x316` | f64 | depth start (mm) | `1.0` |
| `0x31E` | f64 | depth end (mm) | `10.0` |
| `0x326` | f64 | transmit frequency (MHz) | `15.625` |
| `0x32E` | f64 | effective PRF (Hz) | `5500.0` |
| `0x336` | f64 | ADC sampling rate (MHz) | `62.5` |
| `0x346` | u32 | number of plane waves | `11` |
| `0x34A` | f64[11] | plane-wave angles (deg) | `-10, -8, … 10` |
| `0x3C6` | f64[8] | TGC gains | `600.0 ×8` |
| `0x45F` | u64 | acquisition timestamp |  |
| `0x418` | str | sequence name |  |
| `0x42C` | str | project |  |
| `0x43C` | str | subject |  |
| `0x448` | str | session |  |
| `0x454` | str | species |  |
| `0x47F` | str | scan stem |  |
| `0x49E` | str | experimenter |  |
| `0x4A7` | str (hex) | device serial number |  |
| `0x4B9` | str (hex) | hardware label |  |

Strings are length-prefixed.

The TGC gains are always a fixed 8 elements, corresponding to the 8 control points of a TGC (time-gain compensation) curve definable in VSX (Verasonics software).

## Variable-length fields & parsing strategy

The header is a fixed core plus a variable metadata tail. The offsets in the table above are exact for our example file, but several fields are variable-length, so absolute offsets are file-specific. Fortunately every variable field is self-describing—it carries its own size inline—so a one-pass forward walk recovers everything without guessing.

There is no size table up front; the header stores only the totals (fixed core, metadata block, and total header/payload byte counts).

Two classes of variable field:

**1. Length-prefixed strings** — a byte-length prefix sits immediately before the text. Two prefix widths are used in different regions:

| Region | Prefix width | Fields |
|---|---|---|
| probe/config (~offset `0x30C`) | u16 (2 bytes) | probe name |
| provenance block (offset `0x418` → end) | u32 (4 bytes) | sequence, project, subject, session, species, scan stem, experimenter, serial (hex-ascii), hardware (hex-ascii), plus a couple of empty / 1-char fields |

A sequential walk of the provenance block tiles cleanly (`16 → 8 → 0 → 8 → 8 → 3 → 1 …` byte records, including an empty field), confirming the prefix is authoritative.

**2. Count-prefixed numeric arrays** — a preceding integer gives the element count:

| Count field | Array |
|---|---|
| `nTime` = 38 | time coordinates (38 × f64), frame indices (38 × u32) |
| `nPlaneWaves` = 11 | plane-wave angles (11 × f64) |

### Consequences for the loader

- **Payload seek is always safe.** The payload starts at `total_header_bytes` (`0x20`),
  independent of any variable content.
- **The fixed head (before the first variable string) is safe at hardcoded offsets.** Dims, voxel sizes, `dt`, time coordinates, frame indices, the orientation block, the centre/transmit frequencies, and probe pitch/depth/focus all live here—i.e. everything needed for the data array and its geometry.
- **The tail (after the probe name) is position-variable** and must be parsed by walking length/count prefixes from a known anchor, never by absolute offset. This makes the loader robust to different subjects / sequences / probes without knowing string lengths ahead of time.

## Known gaps

- **`physical_to_lab` affine is experimental.** The `0x29C` block is read as a 6DOF probe pose (translation + rotation, the `probeToLab` equivalent) and used to build `affines["physical_to_lab"]`. This is validated on a single near-identity pose; the axis order and Euler convention are assumed. A **multi-pose 3D SCAN v2 file** would confirm it — each pose should carry a different `0x29C` block tracing the robot trajectory. Depth origin is recovered from the depth range; lateral/elevation are centred (symmetric in probe space).
- **Single example file.** All offsets were validated on one 2D acquisition. Multi-pose / multi-block layouts (`npose` or `nblockRepeat` > 1) are inferred, not observed.

Ship this as an explicitly **experimental** v2 path so users can open power-Doppler data today, with a clear upgrade path once Iconeus publishes the official spec.

sdiebolt:
Adds support for binary Iconeus SCAN v2 files to `load_scan`, closing the gap where v2 files were rejected. 

Closes #316.

## What

- `load_scan` now sniffs the format and dispatches to `_load_scan_v1` (HDF5) or `_load_scan_v2` (binary).
- `_load_scan_v2` parses the reverse-engineered header, memory-maps the `float64` payload for lazy Dask access, and returns a DataArray in ConfUSIus `(time, z, y, x)` / `(time, pose, z, y, x)` order with millimetre coordinates.
- Depth origin is recovered from the header depth-range pair via a self-validating span search (falls back to a zero origin); lateral/elevation axes are centred.
- Verified end-to-end on the example file: `(38, 1, 92, 128)`, x ∈ [-6.985, 6.985] mm, depth ∈ [1.0, 9.97] mm, time 0.4–15.2 s.

## Caveats (experimental)

- No `physical_to_lab` affine located yet, so `bps_path` is rejected for v2.
- Multi-pose / multi-block layouts are inferred by analogy with v1, not yet validated against real files.
- Any parse failure is re-raised as an experimental notice pointing to the issue tracker.

### DIFF
--- a/.harper-dictionary.txt
+++ b/.harper-dictionary.txt
@@ -67,5 +67,6 @@ stereotactic
 subspaces
 tSNR
 ty
+unvalidated
 velocimetry
 volumewise
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -33,9 +33,10 @@ Current development version for the next ConfUSIus release.
 
 - [`load_scan`][confusius.io.load_scan] now opens binary Iconeus SCAN v2 files in
   addition to HDF5 SCAN v1 files, detecting the format automatically. SCAN v2 support is
-  experimental: data, timing, voxel spacing, and the depth origin are recovered (lateral
-  and elevation axes are centred on zero), but there is no `physical_to_lab` affine yet
-  and multi-pose layouts are inferred
+  experimental: data, timing, voxel spacing, the depth origin, and provenance
+  (subject/session/project/scan/experimenter, serial number, acquisition datetime) are
+  recovered (lateral and elevation axes are centred on zero), but there is no
+  `physical_to_lab` affine yet and multi-pose layouts are inferred
   ([#316](https://github.com/confusius-tools/confusius/issues/316)).
 - **[Napari plugin]** Added an interactive registration panel for volume alignment in
   napari, including linear and non-linear transforms, progress preview, manual and
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -39,7 +39,8 @@ Current development version for the next ConfUSIus release.
   pitch, focal depth, imaging depth, PRF, plane-wave angles, SVD low cutoff, power-Doppler
   integration window) are recovered (lateral and elevation axes are centred on zero). A
   `physical_to_lab` affine is derived from a header block read as a 6DOF probe pose
-  (experimental, assumed convention). Multi-pose layouts are inferred
+  (experimental, assumed convention), and a BPS sidecar composes a `physical_to_brain`
+  affine as for v1. Multi-pose layouts are inferred
   ([#317](https://github.com/confusius-tools/confusius/pull/317)).
 - **[Napari plugin]** Added an interactive registration panel for volume alignment in
   napari, including linear and non-linear transforms, progress preview, manual and
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -31,6 +31,12 @@ Current development version for the next ConfUSIus release.
 
 ### :sparkles: Enhancements
 
+- [`load_scan`][confusius.io.load_scan] now opens binary Iconeus SCAN v2 files in
+  addition to HDF5 SCAN v1 files, detecting the format automatically. SCAN v2 support is
+  experimental: data, timing, voxel spacing, and the depth origin are recovered (lateral
+  and elevation axes are centred on zero), but there is no `physical_to_lab` affine yet
+  and multi-pose layouts are inferred
+  ([#316](https://github.com/confusius-tools/confusius/issues/316)).
 - **[Napari plugin]** Added an interactive registration panel for volume alignment in
   napari, including linear and non-linear transforms, progress preview, manual and
   automatic initialization, saving/loading transforms, and forward/inverse transform
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -33,10 +33,12 @@ Current development version for the next ConfUSIus release.
 
 - [`load_scan`][confusius.io.load_scan] now opens binary Iconeus SCAN v2 files in
   addition to HDF5 SCAN v1 files, detecting the format automatically. SCAN v2 support is
-  experimental: data, timing, voxel spacing, the depth origin, and provenance
-  (subject/session/project/scan/experimenter, serial number, acquisition datetime) are
-  recovered (lateral and elevation axes are centred on zero), but there is no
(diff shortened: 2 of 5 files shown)


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (approved):
LGTM, although I don't have much to compare and test with. We must iterate with feedbacks from users with this type of scans.

sdiebolt:
> LGTM, although I don't have much to compare and test with. We must iterate with feedbacks from users with this type of scans.

Definitely! That's why I added some useful warnings/errors.

PR #318, opened 2026-07-15
fix(napari): warn on irregular video sync

### LINKED ISSUE #315: warn when overlaying videos on irregularly sampled time axes
When loading a video in the napari plugin, ConfUSIus currently synchronizes it using a linear time transform (scale/translate), which assumes regular sampling. This can silently drift when the selected reference DataArray has a non-regular time axis. We should add a warning in the video panel when the selected reference layer time axis is not regularly sampled beyond a small tolerance, e.g.:

> Reference layer time axis is not regularly sampled; video synchronization uses a linear approximation and may drift.

This would make the current limitation much clearer without changing behavior.

sdiebolt:
## Summary
- show a warning in the video panel when the selected reference layer has irregular time sampling
- keep the warning synced when the reference selection changes
- add napari tests for regular and irregular reference layers

Closes #315

### DIFF
--- a/src/confusius/_napari/_video/_video_panel.py
+++ b/src/confusius/_napari/_video/_video_panel.py
@@ -9,6 +9,8 @@ from typing import TYPE_CHECKING
 import numpy as np
 from napari.utils.notifications import show_error, show_info
 from napari_video.napari_video import VideoReaderNP
+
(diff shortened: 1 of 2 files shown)


FelipeCybis (approved):
LGTM

PR #319, opened 2026-07-15
feat(datasets): add `fetch_khallaf_2026` for naked mole-rat dataset of Khallaf et al. 2026

FelipeCybis:
## Summary

Adds `fetch_khallaf_2026`, a fetcher for the naked mole-rat functional ultrasound dataset from Khallaf et al. (2026), *A queen odour mediates reproductive suppression in a eusocial mammal*, Nature ([10.1038/s41586-026-10772-5](https://doi.org/10.1038/s41586-026-10772-5)). The data is hosted on Edmond (the Max Planck Society's Dataverse) as a single ~19.5 GB zip, organised following BIDS and the proposed fUSI extension BEP-040 (https://edmond.mpg.de/dataset.xhtml?persistentId=doi:10.17617/3.7QCU1F).

The point of interest is that no one ever downloads the whole 19.5 GB archive. Dataverse redirects `/api/access/datafile/{id}` to a presigned object-store URL that honours HTTP Range requests, so the new `_dataverse.py` helpers read the zip's central directory once and then pull only the requested members by range.

## What's included

- `fetch_khallaf_2026`, with `datasets` (`rawdata` / `glm` / `bootstrapping`), `subjects`, `sessions`, `runs`, `reconstruction` (`raw` / `resampled` / `both`), and `sourcedata` filters. Files are cached on first call; later calls with the same `data_dir` return without any network access.
- `sourcedata` (the ~7.8 GB raw Iconeus acquisitions) is off by default, gated only by its own flag and never entity-filtered.
- `refresh=True` re-reads the archive index and reconciles the cache: missing files are downloaded, and cached files whose CRC-32 changed upstream are re-downloaded.
- `_dataverse.py`, a fetcher-agnostic wrapper over `remotezip.RemoteZip` reusable for any Dataverse-hosted stored zip. Covers zip-slip rejection, retry with backoff on transient network errors, and `.part` cleanup so a failed download never leaves a truncated file in the cache.
- Registered in the datasets registry, so `confusius datasets --list` shows it.
- A citation banner printed on fetch (`print_citation=False` silences it), plus a `docs/citing.md` entry with the reference and BibTeX.

## Notes

- New runtime dependency: `remotezip>=0.12.3`.
- The article is in Nature; the dataset itself is CC0 1.0.
- `[tool.codespell]` is added with `ignore-words-list = "hart,portugues"`. Both are author surnames in cited references that codespell reads as typos ("Hart" → "Heart"). `portugues` (BrainGlobe's author in `citing.md`) was already breaking the bare `uv run codespell` that AGENTS.md documents, so this fixes that too.


FelipeCybis:

sdiebolt (approved):
LGTM!

PR #321, opened 2026-07-15
feat(signal): add cosine high-pass filtering

### LINKED ISSUE #167: Add cosine filtering

Now that cosine filtering exist in the GLM module, making it more of a standalone filtering method (so it can also be consumed be `clean`), could be useful.

sdiebolt:
Closes #167

## Summary
- add a standalone `confusius.signal.filter_cosine` high-pass filter based on the GLM cosine drift basis
- let `clean` use cosine filtering via `filter_method="cosine"`
- reuse the shared cosine-regressor builder in both signal cleaning and GLM design tests

### DIFF
--- a/src/confusius/signal/core.py
+++ b/src/confusius/signal/core.py
@@ -122,7 +122,7 @@ def clean(
 
     1. Interpolate censored samples (pre-scrubbing).
     2. Detrend.
-    3. Optional temporal filtering (Butterworth or cosine high-pass).
+    3. Temporal filtering (Butterworth or cosine high-pass).
     4. Censor samples.
     5. Regress confounds.
     6. Standardize.
--- a/src/confusius/signal/core.py
+++ b/src/confusius/signal/core.py
@@ -1,13 +1,11 @@
 """Signal cleaning pipeline for fUSI time series."""
 
-import warnings
 from typing import Literal
 
 import numpy as np
 import xarray as xr
 from xarray.core.types import InterpOptions
 
-from confusius._utils.coordinates import get_coordinate_spacings
 from confusius._utils.filtering import make_cosine_drift_regressors
 from confusius.signal.censor import censor_samples, interpolate_samples
 from confusius.signal.confounds import regress_confounds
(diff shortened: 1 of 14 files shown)


src/confusius/signal/filters.py
```diff
@@ -273,3 +281,83 @@ def filter_butterworth(
...
+            "time grid first."
+        )
+
+    regressors, _ = make_cosine_drift_regressors(
+        signals.sizes["time"],
+        low_cutoff,
+        time_spacing,
+    )
+    projection_matrix = regressors @ np.linalg.pinv(regressors)
+    output_dtype = np.result_type(signals.dtype, np.float64)
```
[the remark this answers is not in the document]
sdiebolt: Good catch. Switched to the low-rank residualization form and renamed the helper to `_residualize_cosine_drift`. I also benchmarked it locally against the dense projection-matrix version before changing it: outputs matched to machine precision (~1e-16 max abs diff across 200-5000 timepoints), while the low-rank form was faster and avoided the O(n_time^2) allocation.

sdiebolt:
@FelipeCybis ready to review!

src/confusius/signal/core.py
```diff
@@ -209,22 +306,16 @@ def clean(
...
-        if (
-            "low_cutoff" in filter_butterworth_kwargs
-            or "high_cutoff" in filter_butterworth_kwargs
-        ):
+    if filter_kwargs:
+        if "low_cutoff" in filter_kwargs or "high_cutoff" in filter_kwargs:
             raise ValueError(
-                "Pass low_pass/high_pass directly to clean, not in "
-                "filter_butterworth_kwargs."
+                "Pass low_pass/high_pass directly to clean, not in filter_kwargs."
```
FelipeCybis:
```suggestion
                "Pass low_cutoff/high_cutoff directly to clean, not in filter_kwargs."
```

src/confusius/signal/filters.py
```diff
@@ -273,3 +283,86 @@ def filter_butterworth(
...
+        kwargs={
+            "regressors": regressors,
+            "pinv_regressors": pinv_regressors,
+        },
+        vectorize=True,
+        dask="parallelized",
+        output_dtypes=[output_dtype],
+    )
+
+    return result.transpose(*signals.dims)
```
FelipeCybis: Can't we just use `regress_confounds` at this point?

src/confusius/signal/core.py
sdiebolt: Yep, fixed the wording to `low_cutoff/high_cutoff`.

src/confusius/signal/filters.py
sdiebolt: Yep, agreed. `filter_cosine` now just builds the cosine drift confounds and delegates the residualization to `regress_confounds(..., standardize_confounds=False)`.

sdiebolt: Indded, would make more sense!

FelipeCybis (approved):
LGTM! Thanks for doing this!

PR #327, opened 2026-07-17
feat(docs): show example summary on gallery card hover

FelipeCybis:
Hovering an example gallery card now fades its first paragraph in over the thumbnail, mirroring the sphinx-gallery/nilearn tooltip.

- The summary is centered over the thumbnail as flowing text (so inline links, abbreviations, and code render inline), behind a near-opaque, theme-aware tint that blurs the thumbnail; the title stays readable below and the thumbnail link stays clickable through the overlay.
- Cross-references and inline links in the summary are flattened to plain text, since they would either not resolve from the index page or render as live links inside an overlay that is inert by design.
- Overlay text is size-matched to the docs' body prose and clamped to the thumbnail height, ellipsizing a long paragraph rather than overflowing.

https://github.com/user-attachments/assets/ad5bc2fc-f56f-48cc-89f2-4f4f2ffa648f

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -133,6 +133,9 @@ Current development version for the next ConfUSIus release.
   `collapse` cell tag, with optional custom title and type (`collapse[<type>]:
   <title>`), i.e. `# %% tags=["collapse[warning]: Collapsed warning"]`
   ([#309](https://github.com/confusius-tools/confusius/pull/309)).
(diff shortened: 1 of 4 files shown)


FelipeCybis:
@sdiebolt quick and good to be reviewed!

sdiebolt (approved):
LGTM, very nice design feature!

PR #334, opened 2026-07-20
feat(decoding): add SearchLight decoder

FelipeCybis:
## Summary

- Added a new `confusius.decoding` module with `SearchLight`, an xarray-native searchlight decoder that maps where a cross-validated scikit-learn estimator can predict a target from each voxel's local neighborhood, keeping fUSI metadata and geometry throughout.
- API reference page, a user-guide stub, and a worked example that decodes locomotion speed from a single fUSI plane and compares the searchlight map against a GLM fit on identically preprocessed data.
- Test suite (brute-force reference, sparse-mask ordering, planted-signal recovery, radius-in-coordinate-units, serial/parallel parity, and input-validation guards) plus changelog and nav entries.

## How SearchLight works

For every voxel selected by `process_mask`, `fit` gathers all `mask` voxels lying within `radius`, clones the estimator onto that neighborhood, cross-validates it against the target over the `time` (sample) axis, and stores the mean score. `radius` is measured in physical coordinate units rather than voxel indices, so anisotropic fUSI voxels give isotropic neighborhoods. The neighborhoods are found once with a KDTree; scoring is parallelized over batches of centers with joblib. Folds are contiguous by default (`KFold(shuffle=False)` for regressors) so that strongly autocorrelated neighboring volumes are not split across the train/test boundary — a `StratifiedKFold` is used for classifiers, and grouped/blocked CV is recommended when class blocks are themselves autocorrelated. The output, `scores_`, is a DataArray in the spatial geometry of `process_mask`, NaN outside it.


FelipeCybis:
@sdiebolt this should be good to be reviewed!

docs/examples/06_decoding/01_searchlight_speed.py
```diff
@@ -0,0 +1,307 @@
...
+# ## Compare the two maps
+#
+# The searchlight reports a cross-validated coefficient of determination, so values at or
+# below zero mean the local neighborhood predicts the speed regressor no better than the
+# fold mean; we clip the color scale at zero. The GLM reports a z-score for the speed
+# contrast. Both maps cover the whole plane.
+
+# %%
+is_dark_theme = sum(mpl.colors.to_rgb(bg_color)) / 3 < 0.5
+searchlight_cmap = "inferno" if is_dark_theme else cc.cm.CET_L17
```
sdiebolt: Rather than adding colorcet as a new dependency, couldn't you just use `inferno_r` in this case?

docs/examples/06_decoding/01_searchlight_speed.py
```diff
@@ -0,0 +1,307 @@
...
+glm_cmap = "berlin" if is_dark_theme else "coolwarm"
+
+background_data = data.mean("time").fusi.scale.db()
+
+fig, axes = plt.subplots(1, 2, figsize=(10, 3), constrained_layout=True)
```
sdiebolt, line 263: Example is missing a thumbnail, this cell should do it?

docs/examples/06_decoding/01_searchlight_speed.py
```diff
@@ -0,0 +1,307 @@
...
+data
+
+# %% [markdown]
+# ## Correct for motion
+#
+# The rat moves freely, so a real analysis should first correct volume-to-volume brain
+# motion. We skip it here to keep the example fast to build, at the cost of a little
+# statistical robustness in the maps below.
+#
+# !!! note "Recommended in a real analysis"
```
sdiebolt, line 67: If having a pre-registered recording becomes a necessity, we could also upload one motion-corrected recording to OSF and use that in our examples, like nilearn!

docs/examples/06_decoding/01_searchlight_speed.py
FelipeCybis: yes, I knew you would have something to say about this and would like your opinion haha. Aesthetically, `inferno_r` starts with yellow, which is not as pleasant as the white in this colorcet colormap. Another argument to defend colorcet is that they have zero dependencies. But I also agree that we should strive for less dependencies.

### `inferno_r`
image

### `CET_L17`
image

FelipeCybis: Looking the two figures now, I don't think `inferno_r` is that bad...

sdiebolt: I see what you mean, colorcet is indeed nicer here... I'm pretty hesitant in the end. On the one hand I love colorcet and use it in pretty much all my figures. On the other hand, it seems simpler in examples to use matplotlib's default color maps, and it indeed avoids a new dependency. Maybe here some thresholding or array alpha would help?

FelipeCybis: I think I am ok with the yellow, in my mind it was worse, but I think I was contrasting with the rest of my screen that is a bit dark. Since the whole page will be in light mode I think we can use `inferno_r`.

docs/examples/06_decoding/01_searchlight_speed.py
FelipeCybis: indeed! I am always forgetting the thumbnail

FelipeCybis: resovled in [4745813](https://github.com/confusius-tools/confusius/pull/334/commits/4745813249ed5645312c0ece8aaa2b270b62e5b0)

docs/examples/06_decoding/01_searchlight_speed.py
FelipeCybis: resolved in [02f8363](https://github.com/confusius-tools/confusius/pull/334/commits/02f8363e14e146b4f1344e5de6fc7fbbf604b29b)

sdiebolt (approved):
LGTM!

PR #343, opened 2026-07-24
fix(io): keep NIfTI and EchoFrame loading lazy under Dask

### LINKED ISSUE #329: fix(nifti): preserve ArrayProxy when constructing lazy Dask arrays
`load_nifti()` currently builds its Dask array from `np.asanyarray(img.dataobj).T`. For an uncompressed NIfTI this is a `numpy.memmap`, but `dask.array.from_array()` (Dask 2026.7.0) copies it while building the graph. Loading many recordings therefore consumes roughly the whole dataset in RSS before any `.compute()`.

Minimal reproducer (256 MiB sparse memmap; `from_array` adds ~512 MiB RSS here):

```python
import os, tempfile
import dask.array as da
import numpy as np
import psutil

with tempfile.TemporaryDirectory() as directory:
    path = os.path.join(directory, "data.bin")
    with open(path, "wb") as file:
        file.truncate(256 * 2**20)
    data = np.memmap(path, dtype=np.float32, mode="r", shape=256 * 2**20 // 4)
    rss = lambda: psutil.Process().memory_info().rss / 2**20
    print(f"memmap RSS: {rss():.0f} MiB")
    _ = da.from_array(data, chunks=2**20, asarray=False, name=False)
    print(f"after from_array RSS: {rss():.0f} MiB")
```

`nibabel.ArrayProxy` itself is sliceable and can be passed directly to Dask. Construct from the proxy, reverse tuple chunk specifications before transposing to ConfUSIus dimension order, and retain the proxy in the Dask graph. A fresh-process check then loads 168 NIfTIs in ~0.1 GiB RSS and processes 1,333 256-volume batches at <0.5 GiB peak.

sdiebolt:
## Summary
- `load_nifti` was materializing nibabel's `ArrayProxy` into a `numpy.memmap` (via `np.asanyarray`) before wrapping it in `da.from_array`, which copies the whole array while building the graph instead of staying lazy. Now passes the `ArrayProxy` straight through and transposes the resulting Dask array afterward instead.
- `load_echoframe_dat` returned a raw `numpy.memmap`; it now builds a lazy `dask.array.Array` via one `dask.delayed` task per acquisition block (Dask's documented mmap-chunking pattern) and assembles it into a ConfUSIus-ordered `(time, z, y, x)` `xarray.DataArray` with coords/attrs, matching `load_nifti`/`load_scan`. `convert_echoframe_dat_to_zarr` keeps its exact public signature but now consumes that DataArray instead of duplicating coordinate/attribute assembly and transpose/reshape logic.

Closes #329

### DIFF
--- a/src/confusius/io/echoframe.py
+++ b/src/confusius/io/echoframe.py
@@ -289,7 +289,7 @@ def _load_echoframe_dat_blocks(
 
 def load_echoframe_dat(
     dat_path: str | Path,
-    meta_path: str | Path,
+    meta_path: str | Path | None = None,
     dat_dtype: npt.DTypeLike = np.complex64,
     header_dtype: npt.DTypeLike = np.uint64,
     n_header_items: int = 5,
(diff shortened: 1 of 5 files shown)


sdiebolt:
@FelipeCybis Ready to review!

FelipeCybis (approved):
LGTM

PR #351, opened 2026-07-31
fix(signal): z-score standardized confounds

### LINKED ISSUE #350: Make confound standardization z-score confounds
ConfUSIus currently treats `standardize_confounds=True` as max-absolute division without centering. That leaves nonzero confound means in the regression design, so confound regression can remove mean-related signal when users expect only confound fluctuations to be removed.

In current Nilearn's confound-regression step, `standardize_confounds=True` z-scores confounds. With `standardize_confounds=False`, Nilearn divides confounds by their maximum absolute value for numerical stability instead of centering them. If detrending/filtering is requested, Nilearn applies that to confounds earlier in the cleaning pipeline.

Align `cf.signal.clean` and `cf.signal.regress_confounds` with the confound-regression behavior and document the distinction.

sdiebolt:
## Summary
- make `standardize_confounds=True` z-score confounds before regression
- keep max-absolute division for `standardize_confounds=False`
- preserve existing cosine-filter behavior with its constant drift regressor

Closes #350.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -10,6 +10,12 @@ icon: lucide/history
 
 Current development version for the next ConfUSIus release.
 
+### :bug: Fixes
+
+- Confound regression now z-scores confounds when `standardize_confounds=True`, so
+  motion-confound cleaning removes fluctuations without regressing baseline-related
+  signal by default ([#351](https://github.com/confusius-tools/confusius/pull/351)).
+
 ## 0.6.0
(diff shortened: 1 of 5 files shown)


sdiebolt:
@FelipeCybis quick one, ready to review!

FelipeCybis (approved):
Nice catch! And thanks for the missing changelog entry as well!

PR #353, opened 2026-08-01
fix(timing): reduce resampling memory

### LINKED ISSUE #328: resample_to_uniform_time uses excessive RAM on large DataArray
## Summary
`resample_to_uniform_time` can use far more memory than the input size on large arrays (for example ~10 GB input -> ~70 GB RAM).

## Likely cause
`resample_time` currently routes through `xarray.apply_ufunc(..., dask="parallelized")` + `scipy.interpolate.interp1d`, which appears to:
- interpolate large non-time blocks in one shot,
- let SciPy copy the input array internally,
- promote `float32` input to `float64` output.

## Expected
Memory use should stay much closer to input + output size, especially for `float32` inputs.

## Possible fixes
- use `interp1d(..., copy=False, assume_sorted=True)`
- avoid unconditional `float64` promotion
- chunk/process over non-time dims for large arrays

sdiebolt:
## Summary
- keep float32 outputs when resampling float32 data
- avoid avoidable scipy interp1d copies
- require sorted time coordinates before using assume_sorted=True

Closes #328


src/confusius/timing.py
```diff
@@ -326,8 +326,15 @@ def resample_time(
     validate_time_series(data, "time resampling")
 
     time_coord = data.coords[TIME_DIM].values
+    if np.any(np.diff(time_coord) <= 0):
+        raise ValueError("time coordinates must be strictly increasing for resampling.")
+
```
FelipeCybis: I would probably see this inside `validate_time_séries` as an optional argument and then transfer the rejection test to the validation module as well. What do you think?

sdiebolt: Done in da53f21

FelipeCybis (approved):

PR #370, opened 2026-08-10
fix(plotting): exclude non-finite values from composite normalization

### LINKED ISSUE #369: `plot_composite` normalization produces NaN with db-scaled data
`plot_composite`/`VolumePlotter.add_composite` produces an all-NaN (blank) composite when either input has been scaled with `.fusi.scale.db()`.

`db_scale` maps zero/negative voxels to `-inf` (common for background/zero voxels in power-Doppler fUSI data). The normalization helper `scale_min_max` (and the `"shared"` branch in `add_composite`) computed `arr.min()`/`arr.max()` directly on the raw array, so `-inf` became the min. This made `(arr - lo) / (hi - lo)` evaluate to `inf/inf = nan` for every pixel, silently blanking the whole composite instead of raising or rendering.

sdiebolt:
## Summary
- `db_scale` maps zero/negative voxels to `-inf`, which `scale_min_max` and the `"shared"` branch of `add_composite` fed straight into `min()`/`max()`, turning every normalized pixel into `nan` and silently blanking `plot_composite` output.
- Both now exclude non-finite values from the normalization bounds and clip output to `[0, 1]`.

Fixes #369

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -40,6 +40,13 @@ Current development version for the next ConfUSIus release.
   `beamforming_sound_velocity` arguments
   ([#322](https://github.com/confusius-tools/confusius/pull/322)).
 
+### :bug: Fixes
+
+- `plot_composite` no longer produces a blank/NaN composite when either input has
+  been scaled with `.fusi.scale.db()`; the `-inf` values `db_scale` assigns to
+  zero-valued voxels are now excluded from the normalization bounds
+  ([#370](https://github.com/confusius-tools/confusius/pull/370)).
+
 ## 0.6.1
 
 Released 2026-08-07.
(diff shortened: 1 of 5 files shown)


sdiebolt:
Addressed both Copilot review comments:

- `scale_min_max`'s docstring overclaimed that non-finite values are "clipped to 0/1" — that's true for `-inf`/`inf` but not for `nan`, since `np.clip(nan, ...)` propagates `nan` (nothing to clamp it to on the `[0, 1]` scale). Reworded to be precise about which values get clipped vs. which remain `nan`.
- Left the `where=`/`initial=` allocation-avoidance suggestion as a follow-up rather than taking it now — real but minor for the composite panel sizes this touches, and I wanted to keep the fix diff-minimal.
- Added tests covering the `ValueError` branches (all-non-finite input) in both `scale_min_max` and the `"shared"` strategy of `add_composite`, closing the two diff-coverage gaps Codecov flagged.

FelipeCybis (approved):
LGTM!

PR #380, opened 2026-08-18
fix(xarray): suppress RuntimeWarning in db_scale/log_scale under Dask

### LINKED ISSUE #379: fix(xarray): db_scale/log_scale leak dask RuntimeWarning for zero/negative values
## Bug

`db_scale`/`log_scale` wrap `xr.ufuncs.log10`/`log` in `np.errstate(divide="ignore", invalid="ignore")`, but for Dask-backed data the log runs lazily on worker threads at `.compute()` time, after the `errstate` context has already exited. Zero/negative inputs then raise:

```
RuntimeWarning: divide by zero encountered in log10
```

## Repro

```python
pwd.mean("time").fusi.scale.db().fusi.plot.volume().show()
```

## Fix

Set `errstate` inside the deferred per-block function (e.g. via `xr.apply_ufunc` with `dask="parallelized"`) instead of around the lazy graph-construction call.

sdiebolt:
## Summary
- `errstate` was scoped around lazy graph construction, not around the deferred `log`/`log10` call Dask actually runs on worker threads at compute time — zero/negative values leaked a `RuntimeWarning`.
- Moved `errstate` inside the per-block function, applied via `xr.apply_ufunc(..., dask="parallelized")`.
- Output dtype now mirrors numpy's own `log`/`log10` promotion (float stays float, int/bool promotes to float64).

Fixes #379


FelipeCybis (approved):
LGTM

PR #398, opened 2026-08-26
feat(signal)!: accept NumPy and DataFrame confounds across the public API

### LINKED ISSUE #384: Allow array-like `sample_mask` and `confounds` in clean

It could be useful to relax clean inputs to accept `NumPy` arrays in addition to `DataArrays` for `sample_mask` and `confounds`. When raw arrays are passed, you could assume that they share data time coordinates?

FelipeCybis:
Closes #384.

## Summary

Every public function that accepts `confounds` now accepts the same three types: `xarray.DataArray`, `numpy.ndarray`, or `pandas.DataFrame`. This applies to `clean`, `regress_confounds`, `FirstLevelModel.fit`, and `make_first_level_design_matrix`. `censor_samples` and `interpolate_samples` accept `sample_mask` as a DataArray or a NumPy array.

## Input rules

- A DataArray must have a `time` dimension. If the DataArray and the signals both have `time` coordinates, they must match.
- A DataFrame must have a `time` column. The function validates this column like DataArray `time` coordinates. The other columns are the confounds, and their names become the `confound` coordinates.
- A NumPy array must have `time` on the first axis and the same length as the signals. The function wraps it with the `time` coordinates of the signals and emits a `UserWarning`, because it cannot verify the alignment. Exception: the GLM does not warn for NumPy input, because a plain matrix is the usual design-matrix input.
- Other input types raise a `TypeError`.

## Changes

- New `confusius.validation.ensure_time_aligned(signals, value, name)` applies the rules above and returns a `(time[, confound])` DataArray. The `allow_dataframe=False` flag rejects DataFrames for `sample_mask`.
- `clean` applies `ensure_time_aligned` to `sample_mask` and `confounds` before the first cleaning step. The warning occurs one time for each NumPy argument. The steps that follow receive DataArrays.
- `censor_samples`, `interpolate_samples`, and `regress_confounds` use `ensure_time_aligned` in their input validation.
- `make_first_level_design_matrix` applies `ensure_time_aligned` to DataFrame and DataArray `confounds` against `volume_times`. Column names come from DataFrame columns or from string DataArray coordinates. Integer coordinates, for example CompCor `component`, get the default `confound_<i>` names as before. `FirstLevelModel.fit` forwards its `confounds` unchanged.

## Breaking change

DataFrame `confounds` passed to the GLM must now have a `time` column. Before this change, the GLM accepted any DataFrame with the correct number of rows.

## Tests

- NumPy inputs give the same result as DataArray inputs with the same values, for `censor_samples`, `interpolate_samples`, `regress_confounds`, and `clean` (Butterworth and cosine paths).
- DataFrame inputs give the same result as DataArray inputs for `regress_confounds` and `clean`.
- `clean` emits exactly one warning for each NumPy argument.
- Wrong length, wrong number of dimensions, unsupported input types, a DataFrame without a `time` column, and a DataFrame with mismatched times raise errors.
- The GLM gives the same design matrix for a DataArray and for a DataFrame, names columns from string coordinates, and rejects mismatched `time` values.


sdiebolt:
@FelipeCybis I think there might be a problem with signals that have `(time, pose)` time coordinate:

```python
In [1]: import numpy as np
   ...: import xarray as xr
   ...: from confusius.validation.time_series import ensure_time_aligned
   ...:
   ...: time = xr.DataArray(np.arange(6).reshape(3, 2), dims=("time", "pose"))
   ...: signals = xr.DataArray(
   ...:     np.zeros((3, 2, 4)), dims=("time", "pose", "space"), coords={"time": time}
   ...: )
   ...: confounds = np.zeros((3, 1))  # plain ndarray, no time coord
   ...:
   ...: result = ensure_time_aligned(signals, confounds, "confounds", ndim=2)
   ...: print("time" in result.coords)  # False, despite the warning claiming it was assigned
<ipython-input-1-055a28cc9091>:11: UserWarning: confounds has no 'time' coordinates, so its alignment with signals cannot be verified; assuming it is ordered like signals along 'time' and using the 'time' coordinates of signals.
  result = ensure_time_aligned(signals, confounds, "confounds", ndim=2)
```

`result` loose its `time` coordinate and `pose` dim!

This could happen to users that would confusingly create per-pose confounds, which isn't supported currently (one would have to filter each pose independently to achieve the same goal). But the error would be confusing in that case, wouldn't it?

FelipeCybis:
> @FelipeCybis I think there might be a problem with signals that have `(time, pose)` time coordinate:
> 
> ```python
> In [1]: import numpy as np
>    ...: import xarray as xr
>    ...: from confusius.validation.time_series import ensure_time_aligned
>    ...:
>    ...: time = xr.DataArray(np.arange(6).reshape(3, 2), dims=("time", "pose"))
>    ...: signals = xr.DataArray(
>    ...:     np.zeros((3, 2, 4)), dims=("time", "pose", "space"), coords={"time": time}
>    ...: )
>    ...: confounds = np.zeros((3, 1))  # plain ndarray, no time coord
>    ...:
>    ...: result = ensure_time_aligned(signals, confounds, "confounds", ndim=2)
>    ...: print("time" in result.coords)  # False, despite the warning claiming it was assigned
> <ipython-input-1-055a28cc9091>:11: UserWarning: confounds has no 'time' coordinates, so its alignment with signals cannot be verified; assuming it is ordered like signals along 'time' and using the 'time' coordinates of signals.
>   result = ensure_time_aligned(signals, confounds, "confounds", ndim=2)
> ```
> 
> `result` loose its `time` coordinate and `pose` dim!
> 
> This could happen to users that would confusingly create per-pose confounds, which isn't supported currently (one would have to filter each pose independently to achieve the same goal). But the error would be confusing in that case, wouldn't it?

Yes, that is tricky. I am proposing that if `signals` has `pose` dimension and confounds are passed, we will warn that all poses are being regressed by the same confounds signals.

In the example you created, a warning saying that no timing is attached should now be the correct behavior.

sdiebolt:
Nice! That's indeed a better guard for users that would inadvertently pass pose-dependent signals. Although we might have the same problem with `interpolate_samples`:

```pycon
>>> import numpy as np, xarray as xr
>>> from confusius.signal.core import clean
>>> time = xr.DataArray(np.tile(np.arange(6.0), (2, 1)).T, dims=("time", "pose"))
>>> signals = xr.DataArray(np.zeros((6, 2, 3)), dims=("time", "pose", "space"), coords={"time": time})
>>> clean(signals, detrend_order=1, sample_mask=np.array([1, 1, 1, 0, 1, 1], dtype=bool))
ValueError: Input DataArray is not 1-D.
```

sdiebolt (approved):
Did a few documentation tweak to clarify, but now LGTM!

PR #399, opened 2026-08-26
fix(nifti): keep the primary header form recoverable after apply

### LINKED ISSUE #397: apply("world_to_sform") on a qform-only NIfTI loses the original qform on save
## Bug

Loading a NIfTI with only a valid qform, adding a sform relationship, applying it, and saving drops the original qform from the written header. The qform is silently replaced (or disabled) instead of being preserved as the secondary form.

## Reproduce

```python
import numpy as np, nibabel as nib
import confusius as cf

qform = np.diag([1.0, 2.0, 3.0, 1.0]); qform[:3, 3] = [10, 20, 30]
img = nib.Nifti1Image(np.zeros((4, 3, 2), np.float32), qform)
img.header.set_qform(qform, code=1)
img.header.set_sform(None, code=0)
img.to_filename("q_only.nii.gz")

da = cf.io.load_nifti("q_only.nii.gz")
world_to_sform = np.eye(4); world_to_sform[0, 1] = 0.3; world_to_sform[:3, 3] = [1, 2, 3]
da.attrs["affines"] = {"world_to_sform": world_to_sform}
da.fusi.affine.apply("world_to_sform", inplace=True)
cf.io.save_nifti(da, "out.nii.gz")

h = nib.load("out.nii.gz").header
print(h.get_qform(coded=True))  # (None, 0): original qform gone
print(h.get_sform(coded=True))  # sheared sform, code 2
```

Saving *without* the `apply` step round-trips correctly (qform code 1 + sform code 2, both matrices right). Only the `apply` path loses the qform.

## Observed

- With a sheared sform: warning "The coordinate-defining affine contains shear... Writing this geometry to sform instead and disabling qform", `qform_code=0`, qform not written.
- With a shear-free sform: no warning, qform and sform are both written with the sform geometry, and qform carries the stale `qform_code=1` from the loaded file. Silent, and the scanner-anat code now labels non-scanner geometry.
- Symmetric case: load a file with both forms, `apply("world_to_qform")`, save: sform is lost the same way.

## Root cause

`load_nifti` stores no `attrs["affines"]` entry for the primary form (it *is* the world frame; see comment in `_resolve_spatial_geometry_from_nifti`). `apply_affine` drops the applied key and re-expresses the remaining keys against the new frame, but has no key for the frame it just left, so after `apply("world_to_sform")` the array holds `affines={}` and no trace of the qform. On save `_prepare_nifti_xforms` falls back to `voxel_to_world` for qform, which is now the sform geometry.

## Proposed fix

Store an identity entry for the primary form on load (`world_to_qform = I` when qform is primary, `world_to_sform = I` when sform is primary). `apply_affine` then re-expresses it to `I @ inv(applied)`, which is exactly the relationship of the old frame to the new one, and `save_nifti` picks it up through its existing default key lookup. No change needed in `apply_affine` or `_prepare_nifti_xforms` for the matrices themselves.

FelipeCybis:
Closes #397

## Problem

Loading a qform-only NIfTI, adding `attrs["affines"]["world_to_sform"]`, applying it with `da.fusi.affine.apply("world_to_sform", inplace=True)` and saving dropped the original qform from the header: `load_nifti` stored no `affines` entry for the primary form, so after `apply` moved the world frame there was nothing left describing where it came from. Depending on shear, the saved file either disabled qform with a warning or silently wrote the sform geometry into qform under the stale `qform_code`. The symmetric case (both forms loaded, `apply("world_to_qform")`) lost the sform the same way.

## Fix

`load_nifti` now stores an identity `world_to_<primary>` entry (`world_to_qform` or `world_to_sform`, whichever defines the world frame). `apply_affine` already re-expresses every stored entry against the new frame, so the identity turns into the exact relationship between the new world frame and the original primary form, and `save_nifti` picks it up through its existing default-key lookup. No change to `apply_affine` or the save path.

The napari B-spline loader strips the new NIfTI-added `world_to_sform` entry the same way it already stripped `world_to_qform`.

## Tests

- New `test_roundtrip_qform_after_applying_added_sform_key`: the issue scenario, asserts qform matrix and code survive and sform equals the applied geometry.
- `test_roundtrip_qform_after_apply_affine_with_singleton_dim` now asserts sform survives as the original sform (previously documented as lost).
- Load tests updated for the identity entry on the primary form.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -148,6 +148,11 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- `load_nifti` now stores an identity `world_to_qform`/`world_to_sform` entry for
+  the coordinate-defining form, so applying another affine (e.g.
+  `.fusi.affine.apply("world_to_sform")`) keeps that form recoverable and
+  `save_nifti` writes it back to the header
+  ([#399](https://github.com/confusius-tools/confusius/pull/399)).
 - `save_nifti` now always writes both a qform and sform (previously sform was
   silently dropped when no secondary affine had been explicitly recorded)
   ([#278](https://github.com/confusius-tools/confusius/pull/278)).
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -148,11 +148,6 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
-- `load_nifti` now stores an identity `world_to_qform`/`world_to_sform` entry for
-  the coordinate-defining form, so applying another affine (e.g.
-  `.fusi.affine.apply("world_to_sform")`) keeps that form recoverable and
-  `save_nifti` writes it back to the header
-  ([#399](https://github.com/confusius-tools/confusius/pull/399)).
 - `save_nifti` now always writes both a qform and sform (previously sform was
   silently dropped when no secondary affine had been explicitly recorded)
   ([#278](https://github.com/confusius-tools/confusius/pull/278)).
--- a/docs/user-guide/voxeldata.md
+++ b/docs/user-guide/voxeldata.md
@@ -384,10 +384,10 @@ pose.
 
 Several loaders populate `.attrs["affines"]` automatically:
 
-- **NIfTI**: The world space is whichever affine (`sform` or `qform`) was selected—it
-  never appears in `.attrs["affines"]` itself. When the other one is also valid, it is
-  stored as `"world_to_sform"` or `"world_to_qform"` accordingly, so the world space can
-  be switched between the two. [`save_nifti`][confusius.io.save_nifti] can write any
+- **NIfTI**: The world space is whichever affine (`sform` or `qform`) was selected; its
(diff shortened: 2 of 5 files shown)


FelipeCybis:
@sdiebolt ok to be reviewed

docs/changelog.md
```diff
@@ -148,6 +148,11 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- `load_nifti` now stores an identity `world_to_qform`/`world_to_sform` entry for
+  the coordinate-defining form, so applying another affine (e.g.
+  `.fusi.affine.apply("world_to_sform")`) keeps that form recoverable and
+  `save_nifti` writes it back to the header
+  ([#399](https://github.com/confusius-tools/confusius/pull/399)).
```
sdiebolt:
```suggestion
```

I'd remove it. We already kept both affines in 0.6.1 I think, so this change is only internal to main.

sdiebolt (approved):

PR #404, opened 2026-08-27
fix(registration): default metric_sampling_percentage to None

sdiebolt:
Change the default because it makes more sense: 100% random sampling isn't equivalent to no sampling, since random sampling is done with replacement.


FelipeCybis (approved):
😅

PR #405, opened 2026-08-27
feat(registration): add intensity_scaling to register_volume/register_volumewise

sdiebolt:
## Summary
- Add `intensity_scaling` (`"none"`/`"db"`/`"sqrt"`/any positive float exponent) to `register_volume` and `register_volumewise`, rescaling only the optimizer's inputs; returned/resampled data keeps original intensities.
- Wire the napari registration panel's existing scale selector to the new argument instead of pre-scaling layers itself.


FelipeCybis (approved):
LGTM!

PR #406, opened 2026-08-27
fix(io): support 4DscanCustom SCAN v1

### LINKED ISSUE #403: `4DscanCustom` Iconeus acquisition mode not supported
### ConfUSIus version

0.6.1

### Python version

3.14

### Operating system

Linux

### Installation

using uv

### Description

Iconeus SCAN files can apparently have `acquisitionMode="4DscanCustom"`, which will make our loader error since we don't expect it. This issue was reported by thisisQ! on Discord. We should fix that by investigating with an actual file.

sdiebolt:
Fixes #403.

- Treat SCAN v1 `4DscanCustom` as the existing 4D layout.
- Keep `iconeus_scan_mode` unchanged for downstream inspection.
- Add a regression test.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -153,6 +153,8 @@ Current development version for the next ConfUSIus release.
 
 ### :bug: Fixes
 
+- `load_scan` now opens Iconeus SCAN v1 files marked as `4DscanCustom`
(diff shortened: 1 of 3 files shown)


FelipeCybis (approved):
LGTM

PR #407, opened 2026-08-27
feat: infer planar slice axis when plotting

### LINKED ISSUE #367: Add a `slice_mode="auto"` mode to `plot_volume`

Currently the default slice mode in `plot_volume` is `"z"`. This forces users to use our `(elevation, depth, lateral)` convention, which is nice because it matches the Allen's, but isn't the most common. It could be great to have an `"auto"` mode that would slice the dimension that has the smallest shape, which most often corresponds to the elevation dimension for linear probes!

sdiebolt:
Fixes https://github.com/confusius-tools/confusius/issues/367.

## Summary
- infer the singleton world dimension when `plot_volume` has no explicit `slice_mode`
- keep singleton world display axes for explicit spatial slicing so line-strip panels plot
- avoid zero-aspect figure sizing crashes
- removed stale plotting examples: to redo later.


FelipeCybis:
I haven't got time to go all the way through this yet, but here is something I found on the napari side if the planar axis is not `z`:

```python
from pathlib import Path
import numpy as np

import confusius as cf

bids_root = cf.datasets.fetch_nunez_elizalde_2022(
    subjects="CR022", sessions="20201007", tasks="spontaneous", acqs="slice02"
)
data_path = (
    Path(bids_root)
    / "sub-CR022"
    / "ses-20201007"
    / "fusi"
    / "sub-CR022_ses-20201007_task-spontaneous_acq-slice02_pwd.nii.gz"
)
volume = cf.load(data_path).mean(dim="time").fusi.scale.db().compute()


transform = np.array([[0, 0, 1, 0], [0, 1, 0, 0], [1, 0, 0, 0], [0, 0, 0, 1]])
volume.affines["world_to_sform"] = transform
volume.fusi.affine.apply("world_to_sform", inplace=True)
volume.fusi.plot()
```

napari plots `z` and some other axis and it does not let us select `xy` as plotting dims.

image

Maybe we can also detect planar slices on the napari side...

sdiebolt:
You can always reorder dims in napari! By right-clicking the dim reordering button. Using your example:

image

sdiebolt:
@FelipeCybis just implemented similar logic for `plot_napari`!

FelipeCybis:
> @FelipeCybis just implemented similar logic for `plot_napari`!

Yes! I forgot that the standard Qt backend for WSL has some problems with mouse dragging and I can't do that, this is not at all a confusius thing nor a napari thing.

FelipeCybis (approved):
Nice!

PR #412, opened 2026-08-28
perf(atlas): batch get_atlas_masks and bump brainglobe-atlasapi to v3

### LINKED ISSUE #345: `brainglobe-atlasapi` V3 compatibility: changes to on-disk format

Hello from the BrainGlobe team!

We're just about to release a new major version of `brainglobe-atlasapi`. There are minimal changes to the Python API, however, the structure of the atlases on disk has been altered significantly. Array-based data is now stored as OME-Zarr, and meshes are in the [neuroglancer precomputed](https://github.com/google/neuroglancer/blob/master/src/datasource/precomputed/meshes.md#multi-resolution-mesh-format) format. Atlas components are now reused across atlases, and fetched lazily. Atlases are now stored in `.brainglobe/brainglobe-atlasapi` by default. Existing atlases (e.g. `.brainglobe/allen_mouse_25um`) can be safely deleted. See the [blog post](https://brainglobe.info/blog/atlas-api-v3.html) for more details. 

You can install the current pre-release version with `pip install brainglobe-atlasapi==3.0.0rc2`.

Please contact us on [Zulip](https://brainglobe.zulipchat.com) if you encounter any issues with the new version!

### LINKED ISSUE #377: `fetch_brainglobe_atlas` eagerly loads and copies atlases
`fetch_brainglobe_atlas("allen_mouse_10um")` currently loads the full atlas into RAM while building the xarray Dataset.

Measured locally with cached `allen_mouse_10um_v1.2`:

| path | RSS after init | RSS after build | dtypes |
| --- | ---: | ---: | --- |
| current without `astype` | 0.31 GB | 8.75 GB | reference uint16, annotation uint32, hemispheres uint8 |
| main with `astype` | 1.52 GB | 20.76 GB | reference float32, annotation int32, hemispheres int8 |

Cause:
- `BrainGlobeAtlas(...)` itself only loads metadata/structures.
- Accessing `atlas.reference`, `atlas.annotation`, `atlas.hemispheres` calls BrainGlobe `read_tiff -> tifffile.imread`, eagerly loading full NumPy arrays.
- Our `astype(...)` calls additionally copy full volumes and change BrainGlobe dtypes.

At minimum, avoid `astype` in the fetcher. Longer term, BrainGlobe v3 should solve some of this.

sdiebolt:
## Summary

- `get_atlas_masks` (`ds.atlas.get_masks(...)`) over many regions was dominated by `xr.concat` recomputing and comparing the full lazily-derived world-coordinate grid across every mask layer, even though all layers share one grid by construction. Skipping that redundant check (`coords="minimal", compat="override"`) is ~125-145x on its own.
- On top of that, batch regions into 8-wide `uint8` bitmask lookups instead of one full `isin` scan of the annotation volume per region (~1.2-2.2x).
- Combined: end-to-end `get_atlas_masks` over 4-128 synthetic regions is **~27-36x faster**, closes #377's underlying complaint about redundant full-resolution work for multi-region calls.
- Bumps `brainglobe-atlasapi` to v3 (closes #345): switches the fetcher to `atlas.template` (not the deprecated `atlas.reference` alias) and `.view()` instead of `.astype()` for `annotation`/`hemispheres`, since v3 already returns same-itemsize unsigned dtypes that only need their sign reinterpreted — avoids a redundant full-array copy on top of BrainGlobe's own eager load (part of #377).
- Fixes a real v3 compat break: BrainGlobe's lazy mesh loader requires `mesh_filename` to be a `pathlib.Path`, not `str`; `structures_from_json` was silently producing the wrong type after a save/load round-trip.
- Updates atlas test fixtures to BrainGlobe v3's Draco-encoded mesh format (was plain OBJ).

Closes #377.
Closes #345.


FelipeCybis (approved):
Noice !

PR #415, opened 2026-08-28
perf(atlas): keep template/annotation lazy as Dask arrays

sdiebolt:
## Summary

Stacked on #412. `BrainGlobeAtlas.template`/`.annotation`/`.hemispheres` (v3 API) always call `.compute()` before returning — so `fetch_brainglobe_atlas` had no way to build a lazy atlas `Dataset`, even though the underlying storage (`ngff_zarr`) is Dask-backed and `create_voxeldata` already accepts `dask.array.Array` directly.

There's no public lazy-loading entry point upstream: [brainglobe/brainglobe-atlasapi#882](https://github.com/brainglobe/brainglobe-atlasapi/issues/882) tracks exactly this ("Load atlas components as lazy dask arrays").

This PR reimplements the path-resolution/download-triggering half of those BrainGlobe properties in `confusius.datasets._brainglobe._load_lazy_ngff_array`, stopping short of `.compute()`. Verified against a real downloaded `allen_mouse_100um` v3 atlas: `reference`/`annotation` come back as genuine `dask.array.Array`, values match BrainGlobe's own eager output exactly, and `validate_atlas` passes. `hemispheres` stays plain numpy for symmetric atlases (BrainGlobe synthesizes it in-memory from `shape`, never reads it from disk, so there's no laziness to preserve there).

**Tradeoff, called out in the docstring**: this depends on `BrainGlobeAtlas` v3 internals (`root_dir`, `metadata`, `fs`, the `_template_pyramid_level`/`_annotation_pyramid_level` attrs) that aren't part of BrainGlobe's public API and already broke once across the v2→v3 migration (#345). I'll comment on brainglobe-atlasapi#882 with our use case to check if we can contribute for a real upstream fix, but wanted to unblock this now.


FelipeCybis (approved):
LGTM

PR #417, opened 2026-08-28
fix(multipose): loosen consolidate_poses regularity check for small steps

### LINKED ISSUE #363: Tolerance in `consolidate_poses` might be too strict
The relative tolerance in `confusius.multipose.consolidate_poses` is currently 1%. When recording angiographies with a 100 um step, this corresponds to a 1 um jitter tolerance, which is very small... I think we should bump the relative tolerance, or set an absolute tolerance?

sdiebolt:
## Summary
- `consolidate_poses` rejected any step-spacing jitter above 1% of the mean spacing, which is unrealistically tight at small step sizes (a 100 um step only tolerated ~1 um of jitter — below real stepper-stage repeatability).
- Adds an `atol` parameter (default 5 um), combined with the existing `rtol` (default 0.01, unchanged) via the standard `abs(spacing - mean_spacing) <= atol + rtol * abs(mean_spacing)` convention, so `atol` dominates at small steps and `rtol` still governs at larger ones.

Closes #363

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -174,9 +174,10 @@ Current development version for the next ConfUSIus release.
 - `consolidate_poses`'s regularity check no longer rejects realistic stage jitter
   on small pose steps (e.g. a 100 um step previously tolerated only ~1 um of
   jitter under a pure 1% relative tolerance). The check now combines a new `atol`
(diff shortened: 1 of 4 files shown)


sdiebolt:
@FelipeCybis ready to review!

FelipeCybis (approved):
LGTM

PR #418, opened 2026-08-28
feat(validation): accept binary numeric masks, not just boolean/integer

### LINKED ISSUE #382: Support binary masks with non-boolean dtypes

`extract_with_mask` currently expects a boolean mask, but some neuroimaging tools (e.g. FSL) may save masks as integer or float arrays.

Would it make sense to accept non-boolean masks when they are clearly binary, and convert them internally to boolean?

This would improve interoperability while still avoiding silently treating arbitrary images/volumes as masks!

sdiebolt:
## Summary
- `validate_mask`/`extract_with_mask` (and the shared dual-input mask helpers) required strict boolean dtype, or integer dtype with a single non-zero value — rejecting binary masks saved as float (e.g. by FSL/NiBabel).
- `check_mask_dtype` now accepts any numeric dtype (int or float) whose non-zero values are all equal to a single value, coercing to boolean internally. More than one distinct non-zero value, or a non-numeric dtype, still raises.

Closes #382


FelipeCybis (approved):
LGTM

PR #419, opened 2026-08-28
fix(registration): cap default slice count in registration progress plot

### LINKED ISSUE #368: `register_volume` progress plotter should not plot all slices
When registering angiographies with many slices, the default progress plotter from `register_volume` becomes pretty slow and difficult to read. We should limit the number of slices by default, and possibly expose arguments from `plot_composite` for users to tweak?

sdiebolt:
## Summary
- `register_volume`'s live progress plot rendered every slice of the composite fixed/moving mosaic, which was slow and hard to read for volumes with many slices.
- `MatplotlibRegistrationProgressPlotter` now caps the mosaic at 9 evenly spaced slices by default (a 3x3 grid), exposed as a new `max_composite_slices` parameter on `register_volume` (`None` restores plotting every slice). The napari progress factory gained the same parameter, kept inert, for call-signature compatibility.

Closes #368


FelipeCybis (approved):
All good once tests are green!

PR #429, opened 2026-08-31
feat(napari): responsive mouse source + pinnable signals

### LINKED ISSUE #428: Signal plotter: live signals can be slow and aren't persistent across source modes
The napari Signal Plotter's mouse-source mode has a few gaps that make it hard to use for real interactive exploration:

- Shift+hover only plots a signal while the mouse is *moving*; pressing Shift while already resting on a voxel does nothing.
- On larger recordings, mouse-driven updates lag and feel unresponsive/jittery rather than tracking the cursor live.
- A live signal (mouse voxel, a point, or a label region) only exists for as long as its source mode is active — switching source modes drops it, with no way to keep a signal around for later use (e.g. as a reference while exploring a different source, or as an eventual confound/seed elsewhere).
- The `ImportedSignal` naming in the signal store only reflects file imports, which doesn't fit signals that originate from a live source.

Relates to the seed-based connectivity exploration design discussed for the plugin (a live-updating seed map needs a responsive, persistable signal-selection UI to build on).

sdiebolt:
## Summary
- Throttle mouse-driven signal-plotter updates (leading-edge, no added latency in the common case) and let Shift-press-without-moving plot the resting cursor's signal.
- Rename `ImportedSignal`/`imported_signals` to `StoredSignal`/`stored_signals` across the signal store, manager, and plotter, since the store no longer only holds file imports.
- Add `SignalStore.pin_signal` + a "Pin" button on the plotter to snapshot a currently plotted live signal (mouse voxel, point, or label region) so it survives a source-mode switch instead of being dropped.
- Add an "Export Selected" button to the Signal Manager dialog.

Closes #428.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,26 +15,20 @@ Current development version for the next ConfUSIus release.
 - Scrolling the sidebar with the mouse wheel no longer gets hijacked by whichever
   combo box or spin box the cursor happens to be over
   ([#431](https://github.com/confusius-tools/confusius/pull/431)).
-
-### :sparkles: Enhancements
-
-- **napari plugin:** The signal plotter's mouse source now updates when Shift is
-  pressed while the cursor is already resting on a voxel, not only while moving;
-  a live signal (mouse voxel, a point, or a label region) can now be pinned so it
-  persists across source-mode switches instead of being dropped, reusable
-  wherever stored signals are — imported and pinned signals now share one
-  concept (`StoredSignal`) throughout the signal store, plotter, and manager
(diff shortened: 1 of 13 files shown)


FelipeCybis (approved):
https://github.com/confusius-tools/confusius/pull/429/commits/5dca31e07239cfc05c0844f38ceb3ad9dadfc185 Fixes some small bugs I found:

- If the signal was pinned while z-scored, it would think the z-score is the raw signal and would not go back to "un"z-scored signal.
- Points colors and signals plot colors were a bit desynchronized when changing the colors from the napari controls.
- DVARS now plots in the main signal plotter.

PR #434, opened 2026-09-01
perf(signal): avoid full SVD in compute_compcor_confounds

### LINKED ISSUE #433: compute_compcor_confounds: unconditional full SVD is slow for typical CompCor shapes
CompCor's component extraction always runs a full SVD over the selected `(time, voxels)` matrix, even though only a handful of components (5-30) are ever kept, and one side of the matrix (usually the voxel count) can be tens of thousands. sklearn's `PCA` picks between an eigh-on-the-small-covariance solver and a randomized solver based on shape and how small `n_components` is relative to the dimensions (see `sklearn.decomposition.PCA`'s `svd_solver="auto"` docs) — a similar heuristic here could avoid computing far more of the spectrum than is actually used.

sdiebolt:
CompCor keeps only a handful of components (5-30) out of what can be thousands of timepoints and tens of thousands of selected voxels, so computing the full spectrum was doing far more work than needed.

Eigendecompose the smaller of the two Gram matrices instead of the full matrix (generalizes nilearn's time-much-less-than-voxels assumption to whichever side is smaller), and switch to randomized SVD above a size/n_components threshold mirroring `sklearn.decomposition.PCA`'s own solver heuristic. Dask-backed inputs get the equivalent `dask.array.linalg.svd_compressed` treatment.

Closes #433.


sdiebolt:
@FelipeCybis ready to review!

FelipeCybis (approved):
LGTM

PR #436, opened 2026-09-01
feat(registration): accept any fixed volume in register_volumewise

FelipeCybis:
## Summary

`register_volumewise` and `data.fusi.register.volumewise` gain a `fixed` keyword: a spatial-only VoxelData array on the same voxel grid as `data` (for example the mean of a few low-motion frames) that every frame is registered to, as an alternative to `reference_time`.

`intensity_scaling` is split into `moving_intensity_scaling` (applied to every frame, same default as before) and `fixed_intensity_scaling` (applied to `fixed` only, defaults to `moving_intensity_scaling`) — a breaking rename.

Napari UI exposing `fixed`/`fixed_intensity_scaling` in the registration panel follows in #437, stacked on this PR.

Closes #376

Issue #376, opened 2026-08-14
Allow any fixed volume in `register_volumewise`


sdiebolt:
### Describe the new feature you'd like

Like in ANTsPy, it could be great to allow passing any fixed reference volume (e.g. computed manually from an average of low intensity frames) to `register_volumewise`, instead of forcing users to use a reference time index.

PR #437, opened 2026-09-01
feat(napari): register within-scan frames to a fixed layer

FelipeCybis:
## Summary

The registration panel's within-scan mode gains a **Use a fixed layer** toggle: instead of a reference volume index, every frame can be registered to a spatial-only layer on the same grid (for example the mean of a few low-motion frames), with its own **Fixed intensity scaling**. This reuses the between-scan mode's existing fixed-layer picker and fixed-scale combo, wiring them to `register_volumewise`'s new `fixed`/`fixed_intensity_scaling` parameters.

The result layer's metadata records `fixed_layer_name`, and `reference_time` is `None` when a fixed layer was used. The moving scaling label now always reads "Moving intensity scaling" in both modes.

Stacked on #436 (the `register_volumewise` API change); this PR's own diff is the napari UI only.

Part of #376.

## Test plan
- [x] `uv run pytest tests/unit/test_napari/test_registration_panel.py`
- [x] `just pre-commit`

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -42,6 +42,9 @@ Current development version for the next ConfUSIus release.
 
 ### :frame_photo: Napari plugin
 
+- The registration panel's within-scan mode can register every frame to a fixed
+  layer, with its own intensity scaling, as an alternative to the reference volume
+  index ([#376](https://github.com/confusius-tools/confusius/issues/376)).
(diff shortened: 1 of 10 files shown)


FelipeCybis:

sdiebolt (approved):
LGTM

PR #440, opened 2026-09-01
feat(registration)!: replace joblib with Dask in register_volumewise

### LINKED ISSUE #439: Lazy per-frame reads of gzip NIfTI re-decompress on every call
Lazy per-frame reads of a gzip-compressed NIfTI (`.nii.gz`) re-decompress on every call, since `cf.load()` wraps it as a lazily dask-backed array (`nibabel.arrayproxy.ArrayProxy` via `da.from_array`) and:

- gzip has no random access — reaching any byte offset requires decompressing sequentially from the start of the file.
- Dask does not cache results across separate `.compute()`/`.values()` calls, so each independent lazy read re-triggers that full decompression — even for two frames in the *same* dask chunk.

Measured on a real recording (1800 frames, ~196MB `.nii.gz`, default time-chunking `(1170, 630)`):

```
frame 0    (cold):             0.89s
frame 1    (cold, same chunk): 0.69s   # no reuse despite being in the same chunk
frame 0    (after .persist()): 0.002s
frame 1799 (after .persist()): 0.002s
```

A per-frame loop over all 1800 frames without persisting first ran for 12+ minutes before being killed; the same loop over 300 frames with `.persist()` called once beforehand took ~16s.

A single full read (`cf.load(path).compute()`, as the docs examples already do) is unaffected — it pays this cost exactly once either way. The gap is specifically "many independent partial reads against an unpersisted gzip-backed array," e.g. any per-frame Dask pipeline (`dask.delayed`/`apply_ufunc` over `time`) that doesn't `.persist()` first.

Suggest documenting this (docstring note on `load_nifti`/`cf.load`, maybe a user-guide callout) recommending `.persist()` before per-frame/chunk loops over lazy gzip-backed data — not disabling lazy loading of `.nii.gz`, which would remove a working capability for the common one-shot-read case.

Found while prototyping a Dask-based `register_volumewise` backend.

sdiebolt:
Replaces joblib with a Dask `distributed.Client` in `register_volumewise`, dropping the eager full-recording materialization (`arr = data_moved.values`) that forced everything into RAM regardless of `n_jobs`. Each frame is now read lazily via `isel(time=t)` as it's dispatched to the Client (ambient if one is active, otherwise a local one created automatically — one worker process per CPU).

**Breaking changes:**
- `n_jobs` is gone — parallelism is controlled by the Dask `distributed.Client` instead.
- `abort_event` must now be a `distributed.Event`, not a `threading.Event`: `Client.submit` always pickles the submitted call (even for a thread-based/in-process worker), and a `threading.Event` can't be pickled or stay live across that boundary — `distributed.Event`, backed by the scheduler, does. `register_volume` accepts either through a new structural `AbortEvent` protocol, since it's shared between napari's single-frame path (`threading.Event`) and `register_volumewise`'s frames (`distributed.Event`).
- h5py-backed (SCAN) input now always raises `TypeError` (previously only when `n_jobs != 1`), since it can never be pickled to a Dask worker. Call `.compute()` first.

**Added:** progress reporting (progress bar + `progress_reporter` callbacks) on the new Client via `distributed.as_completed`, matching the previous joblib-based behavior.

**napari plugin:** the "Parallel jobs" control is now "Parallel workers", sizing a `distributed.Client` the panel creates and owns per run (closed in `_end_work()`) instead of a joblib `n_jobs` value.

Closes #439


FelipeCybis:
Hey! You may want to stack this over #436 and #437. I don't think you will have much of a problem, but might be worth it.

PR #441, opened 2026-09-01
docs(io): note gzip NIfTI lazy-read costs

### LINKED ISSUE #439: Lazy per-frame reads of gzip NIfTI re-decompress on every call

sdiebolt:
Documents the `.nii.gz` lazy-read caveat and when to use `.compute()` vs `.persist()`.

Closes #439.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,12 @@ Current development version for the next ConfUSIus release.
   large recordings or broad noise masks
   ([#434](https://github.com/confusius-tools/confusius/pull/434)).
 
+### :books: Documentation
+
+- Clarified when to use `.compute()` or `.persist()` before repeated partial reads from
+  gzip-compressed NIfTI files
+  ([#441](https://github.com/confusius-tools/confusius/pull/441)).
+
 ### :frame_photo: Napari plugin
 
 - Scrolling the sidebar with the mouse wheel no longer gets hijacked by whichever
--- a/docs/user-guide/io.md
+++ b/docs/user-guide/io.md
@@ -127,6 +127,17 @@ an operation requires it.
     da = cf.load("sub-01_task-awake_pwd.nii.gz").compute()
     ```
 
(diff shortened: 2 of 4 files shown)


sdiebolt:
@FelipeCybis ready to review!

FelipeCybis (approved):

PR #442, opened 2026-09-01
perf(glm): make GLM fit faster

### LINKED ISSUE #310: GLM fit with several runs should be done in parallel

We should add a `n_jobs` parameter to the GLM api so we can run fits of more than one `run` through joblib.

FelipeCybis:
In my machine, `docs/examples/06_glm/01_first_level.py` fits in about 55 seconds on this branch, against about 129 seconds on main. The larger part of the gain applies to the default `noise_model="ar1"`. The z-maps are unchanged, except on voxels where rounding controls the value in both versions.

Closes #310.

### 1. `inv` in place of `pinv` for the per-voxel normal equations

`ARModel.fit` inverts a `(147k, 15, 15)` stack. The pseudo-inverse is a batched SVD and costs about 6 times more than the LU inverse. The two results agree to 1e-14. `pinv` stays for a rank-deficient design, where `inv` returns a wrong answer without an error. The full-rank branch needs no guard of its own. Whitening multiplies the design by a unit lower triangular factor, so it preserves rank for every rho and `XtX` stays positive definite.

Before:

```python
cov_beta = np.linalg.pinv(XtX)  # (V, K, K)
```

After:

```python
if self.df_model == K:
    cov_beta = np.linalg.inv(XtX)  # (V, K, K)
else:
    cov_beta = np.linalg.pinv(XtX)  # (V, K, K)
```

### 2. No second pass over the OLS residuals

The `residuals` property returns `self.Y - self.predicted`, which allocates a second `(n_volumes, n_voxels)` array. OLS whitening is the identity, so `whitened_residuals` already holds those values.

Before:

```python
ols_results = ols_model.fit(data_2d)
rho_per_voxel, _ = estimate_ar_coeffs(ols_results.residuals, order=ar_order)
```

After:

```python
ols_results = ols_model.fit(data_2d)
ols_residuals = cast("npt.NDArray[np.floating]", ols_results.whitened_residuals)
rho_per_voxel, _ = estimate_ar_coeffs(ols_residuals, order=ar_order)
```

### 3. `optimize=True` on two `numpy.einsum` calls

The lagged cross-product goes to BLAS and runs about 30 times faster. The three-operand term contracts in pairs, in place of building a full `(order, order, V, K, K)` intermediate. The two-operand term between them is faster without the flag, so it keeps the default.

Before:

```python
XtX = (
    A[np.newaxis]
    - np.einsum("iv,ikl->vkl", self.rho, B_sym)
    + np.einsum("iv,jv,ijkl->vkl", self.rho, self.rho, C)
)  # (V, K, K)
...
Q = np.einsum("itk,tv->ikv", L, whitened_Y)
```

After:

```python
XtX = (
    A[np.newaxis]
    - np.einsum("iv,ikl->vkl", self.rho, B_sym)
    + np.einsum("iv,jv,ijkl->vkl", self.rho, self.rho, C, optimize=True)
)  # (V, K, K)
...
Q = np.einsum("itk,tv->ikv", L, whitened_Y, optimize=True)
```

## Two smaller items in this PR

`FirstLevelModel` accepts `show_progress`, which shows a progress bar over the runs (defaults to `False`). The bar lives in `confusius/_utils/progress.py`, and `decoding/searchlight.py` now draws its own bar from that helper.

```python
model = FirstLevelModel(noise_model="ar1", show_progress=True)
```

`compute_contrast` no longer prints a divide-by-zero `RuntimeWarning`. A recording that holds a voxel of no variance over time triggered this. An atlas grid leaves whole regions like that outside the recorded field of view.

Before:

```python
return np.where(x == 0, 0, 1.0 / x)
```

After:

```python
return np.divide(1.0, x, out=np.zeros_like(x), where=x != 0)
```


FelipeCybis:
@sdiebolt this should be good to review (#445 to be merged before this one)

PR #445, opened 2026-09-01
fix(plotting): honor a lone `vmin` or `vmax` under `auto_range`

### LINKED ISSUE #443: `plot_stat_map` and `plot_matrix` ignore `vmin` or `vmax` when only one of the two is given
With `auto_range=True` (the default), an explicitly passed `vmin` or `vmax` has no effect on a data that holds both positive and negative values. The colorbar takes its range from the data (passing both `vmin` and `vmax` works).

`_resolve_stat_map_style` in `src/confusius/plotting/image.py` fills the bound that the caller left out from the data, and then takes the larger magnitude of the two:

```python
resolved_vmin = vmin if vmin is not None else data_min
resolved_vmax = vmax if vmax is not None else data_max

if data_min < 0 < data_max:
    abs_max = max(abs(resolved_vmin), abs(resolved_vmax))
    return -abs_max, abs_max
```
On data that spans -30 to 30, `vmax=10` gives `abs_max = max(30, 10) = 30`. The data-derived fallback wins over the explicit bound.

`_resolve_matrix_style` in `src/confusius/plotting/matrix.py` holds the same code and behaves the same way.

Minimal reproducible example
```python
import numpy as np

import confusius as cf

data = np.linspace(-30.0, 30.0, 1000).reshape(10, 10, 10)
stat_map = cf.create_voxeldata(data, dims=("k", "j", "i"), voxel_to_world=np.eye(4))

for kwargs in ({"vmax": 10}, {"vmin": -10}, {"vmin": -10, "vmax": 10}):
    plotter = cf.plotting.plot_stat_map(stat_map, **kwargs)
    norm = plotter.figure.axes[0].collections[0].norm
    print(kwargs, "->", (norm.vmin, norm.vmax))
    plotter.close()

matrix = np.array([[30.0, -30.0], [-30.0, 30.0]])
for kwargs in ({"vmax": 10}, {"vmin": -10}, {"vmin": -10, "vmax": 10}):
    figure, _ = cf.plotting.plot_matrix(matrix, **kwargs)
    norm = figure.axes[0].collections[0].norm
    print(kwargs, "->", (norm.vmin, norm.vmax))
```
Output:
```
{'vmax': 10} -> (-30.0, 30.0)
{'vmin': -10} -> (-30.0, 30.0)
{'vmin': -10, 'vmax': 10} -> (-10.0, 10.0)
{'vmax': 10} -> (-30.0, 30.0)
{'vmin': -10} -> (-30.0, 30.0)
{'vmin': -10, 'vmax': 10} -> (-10.0, 10.0)
```
### Expected behavior

An explicit `vmin` or `vmax` with `auto_range=True` should set both extremes. `vmax=10` gives `(-10, 10)`, and so does `vmin=-10`. The range stays symmetric around zero, which is what `auto_range=True` promises for a diverging map. Only a call that gives neither limit reads the range from the data.

### Notes

The docstring of `plot_stat_map` names two cases where a bound is ignored. Both are about the sign of the data. `vmin` is ignored on a map of non-negative values. `vmax` is ignored on a map of non-positive values. The case in this issue is not one of them.

`auto_range=False` respects a lone `vmax`, and returns (-30, 10) for the example above. That is the documented behavior of the flag, because it drops the anchor at zero.

I found this in `docs/examples/06_glm/01_first_level.py`, which passes `vmax=10` to `plot_stat_map` and gets a colorbar that runs from -30 to 30 (not in the current rendered example because of cache).

FelipeCybis:
Closes #443.

## Fix

With `auto_range=True` and data spanning both signs, the symmetric range is now built only from the bounds the caller actually passed: a lone `vmin` or `vmax` sets the range to `[-|bound|, |bound|]`. The data's own magnitude is used only when neither bound is given.

## Also in this PR

`_resolve_stat_map_style` and `_resolve_matrix_style` were identical apart from their input type and their colormap constant names, whose values were the same three strings. They are now a single `_resolve_colormap_style` in `confusius/plotting/_utils.py` with one set of colormap constants, so the two paths can no longer drift (net 51 lines removed).

## Behavior change

Code that passed a lone `vmin` or `vmax` on sign-spanning data and relied on it being ignored will see a different colormap range. `docs/examples/05_connectivity/02_atlas_seed_map.py` and `docs/examples/06_glm/01_first_level.py` both pass a lone `vmax` and now get the range their prose already claimed ("fixing the colormap to a shared range"), so only the CI-built gallery images change.

A lone `vmin=0` or `vmax=0` on sign-spanning data now resolves to a degenerate `[0, 0]` range, which matplotlib expands to `[-0.1, 0.1]`. That input asks for a symmetric range capped at magnitude zero, so it is left as is rather than falling back to the data, which is exactly the silent-drop behavior this PR removes.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -23,7 +23,7 @@ Current development version for the next ConfUSIus release.
   [`plot_matrix`][confusius.plotting.plot_matrix] now honor a `vmin` or `vmax` passed
   on its own under `auto_range=True`: a lone bound sets the symmetric range to
   `[-|bound|, |bound|]`
-  ([#443](https://github.com/confusius-tools/confusius/pull/443)).
+  ([#445](https://github.com/confusius-tools/confusius/pull/445)).
 
 ### :books: Documentation
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,14 @@ Current development version for the next ConfUSIus release.
   large recordings or broad noise masks
   ([#434](https://github.com/confusius-tools/confusius/pull/434)).
 
+### :bug: Fixes
+
+- [`plot_stat_map`][confusius.plotting.plot_stat_map] and
+  [`plot_matrix`][confusius.plotting.plot_matrix] now honor a `vmin` or `vmax` passed
+  on its own under `auto_range=True`: a lone bound sets the symmetric range to
+  `[-|bound|, |bound|]`
+  ([#443](https://github.com/confusius-tools/confusius/pull/443)).
+
 ### :books: Documentation
 
 - Clarified when to use `.compute()` or `.persist()` before repeated partial reads from
--- a/docs/user-guide/visualization.md
+++ b/docs/user-guide/visualization.md
@@ -438,7 +438,8 @@ By default, `vmin` and `vmax` are taken from the full `stat_map`, and `auto_rang
 picks a suitable range + default colormap automatically:
 
 - Both positive and negative values: diverging, symmetric `[-m, m]` range
-  (`m = max(|vmin|, |vmax|)`), with `cmap="coolwarm"` by default.
+  (`m = max(|vmin|, |vmax|)` over the bounds you pass, or the largest magnitude in
+  `stat_map` when you pass neither), with `cmap="coolwarm"` by default.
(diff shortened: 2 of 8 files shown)


FelipeCybis:

sdiebolt (approved):
LGTM!

PR #446, opened 2026-09-01
feat(xarray): add `get_bounding_box` and speed up atlas mesh clipping

### LINKED ISSUE #444: `get_atlas_mesh` clip step computes the full z/y/x coordinate arrays
## Problem

`get_atlas_mesh` with `clip=True` (default) is slow.

The clip step calls `_drop_vertices_outside_grid` (`src/confusius/atlas/_world_to_base_transform.py`). This helper reads `reference.coords[dim].values` for `z`, `y`, and `x`. For an oblique affine, these world coordinates are lazy `VoxelToWorldTransform` coordinates. Each read computes the full coordinate array for the whole voxel grid. The helper then keeps only the minimum and the maximum of each axis (6 scalars).

For the Allen 25 µm grid (about 77 million voxels), the three reads allocate gigabytes and take seconds. Axis-aligned atlases have eager 1D coordinates, thus they do not show the problem.

## Cause

The world coordinates are an affine function of the voxel coordinates. The helper does not need the full arrays to get the bounds of the grid. The bounds of an affine image of a box are at the 8 corners of the box.

## Possible fix

Add a bounding-box helper that maps the 8 corners of the voxel grid through the voxel-to-world affine. Take the per-axis minimum and maximum of the mapped corners. This gives identical bounds in O(1) instead of O(n_voxels). Then make `_drop_vertices_outside_grid` use this helper.

FelipeCybis:
## Summary

`get_atlas_mesh` with `clip=True` materialized the full lazy `z`/`y`/`x` coordinate grids of an atlas, only to reduce them to per-axis min/max (gigabytes and seconds on Allen-scale grids). Closes #444.

- New public `get_bounding_box(data)` in `confusius.xarray`: maps the 8 corners of the voxel-coordinate box through the voxel-to-world affine. The bounds are identical to the full-grid reduction (the extremes of an affine function over a box are at its corners) and cost O(1) instead of O(n_voxels). The output is a `(bound, component)` DataArray with a `units` attr, and one bounding box per pose for pose-dependent geometry. The input is normalized with `ensure_voxeldata`, so scalar-indexed voxel dims are restored before the computation.
- `_drop_vertices_outside_grid` now clips mesh vertices with these bounds. The one-voxel spacing margin is unchanged, so clipping results are identical.

## Timing

Wall time of `get_atlas_mesh(ds, "root")` on `allen_mouse_25um` (528 x 320 x 456 voxels), best of 3 runs after a warm-up call, on an i7-12700H (WSL2):

| Case | Before (8e48b52) | After | Speedup |
|---|---|---|---|
| Fetched atlas (axis-aligned affine) | 4.91 s | 0.007 s | ~700x |
| World rotated 10 degrees (oblique, post-registration geometry) | 5.08 s | 0.007 s | ~700x |

## Tests

- New `tests/unit/test_xarray/test_affine.py`: oblique, axis-aligned, irregular-coordinate, singleton-dim, and per-pose cases, each compared with a brute-force reference that reduces the materialized world coordinates; scalar-`isel` canonicalization; error paths for non-VoxelData and non-DataArray input.
- Full suite: 2688 passed.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -14,13 +14,13 @@ Current development version for the next ConfUSIus release.
 
 - New [`get_bounding_box`][confusius.xarray.get_bounding_box] returning a VoxelData
   array's world-space bounding box, one per pose for pose-dependent geometry
-  ([#445](https://github.com/confusius-tools/confusius/pull/445)).
+  ([#446](https://github.com/confusius-tools/confusius/pull/446)).
 
 ### :zap: Performance
 
 - [`get_atlas_mesh`][confusius.atlas.get_atlas_mesh] with `clip=True` no longer
   materializes the full world-coordinate grid of an oblique atlas
-  ([#445](https://github.com/confusius-tools/confusius/pull/445)).
+  ([#446](https://github.com/confusius-tools/confusius/pull/446)).
 - [`compute_compcor_confounds`][confusius.signal.compute_compcor_confounds] no
   longer computes a full SVD, extracting components several times faster on
   large recordings or broad noise masks
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,8 +12,10 @@ Current development version for the next ConfUSIus release.
 
 ### :sparkles: Enhancements
 
-- New [`get_bounding_box`][confusius.xarray.get_bounding_box] returning a VoxelData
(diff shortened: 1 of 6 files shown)


FelipeCybis:
@sdiebolt good to be reviewed

PR #448, opened 2026-09-02
feat(atlas): multi-region get_mesh and batched atlas mesh prefetch

FelipeCybis:
Stacked on #446; #263 stacks on top of this one.

## What

- `get_mesh` / `get_atlas_mesh` now take `regions` and `sides` like `get_masks` (one region or a sequence, one side for all or one per region) and return a `{acronym: (vertices, faces)}` dict, with `_L`/`_R`-suffixed keys for single-hemisphere requests. All vertices go through `world_to_base` in one batched pass.
- `fetch_brainglobe_atlas` prefetches every region mesh on the first fetch of an atlas. BrainGlobe 3 downloads meshes lazily, one S3 `exists` + `get` round trip per region (about 1 s each). We list the remote mesh directory once and pull every missing mesh in a single concurrent `s3fs.get` call: the 839 Allen mouse meshes take about 20 s instead of about 14 min, and `get_mesh` works offline afterwards.

## Notes

- Regions with no remote mesh (e.g. RSPd4 in the Allen mouse atlas, brainglobe/brainglobe-atlasapi#647) and an unreachable bucket both warn and fall back to BrainGlobe's lazy per-region download.
- A failed batch removes any partially written file, as BrainGlobe does, since it treats any existing file as a valid cached mesh.
- The remote mesh location is built from `remote_url_s3` and `V3_MESHES_DIRECTORY` in `brainglobe_atlasapi.descriptors`. Upstream has no bulk download API yet (brainglobe/brainglobe-atlasapi#784).
- The atlas test fixtures move to `tests/unit/conftest.py` so other test modules can reuse the mock atlas.


FelipeCybis:
@sdiebolt this should be good to be reviewed!

PR #463, opened 2026-09-18
fix(xarray): keep the applied key as identity in apply_affine

### LINKED ISSUE #461: `apply_affine` removes the applied key from `attrs["affines"]`
## Problem

Since #399, `load_nifti` stores the primary form as an identity entry in `attrs["affines"]` (`world_to_sform` or `world_to_qform`). That entry tells you which space the world coordinates are in.

`apply_affine` should keep this up, but doesn't. `da.fusi.affine.apply("world_to_qform")` makes the `qform` space the new world space, but it silently removes `world_to_qform` from `attrs["affines"]` instead of leaving it as identity. The primary form has an identity entry after a load and no entry after an apply (this was misleadingly not handled in #399)

The result also depends on how you pass the affine:
```python
da.attrs["affines"] = {"world_to_sform": np.eye(4), "world_to_lab": rot}

list(da.fusi.affine.apply("world_to_lab").attrs["affines"])
# ['world_to_sform']

list(da.fusi.affine.apply(rot).attrs["affines"])
# ['world_to_sform', 'world_to_lab'], and 'world_to_lab' is the identity
```
Both calls move the world space the same way. Only the call with the matrix keeps the entry.

### Expected behavior

`apply_affine` never removes entries from `attrs["affines"]`. After `apply("<key>")`, `attrs["affines"]["<key>"]` is the identity, and all other entries are re-expressed in the new world space as they are today.

FelipeCybis:
- Closes #461

## Problem

`da.fusi.affine.apply("<key>")` removed `<key>` from `da.attrs["affines"]`, while applying the same matrix as an array kept the entry as the identity. Since #399, `load_nifti` stores the primary form as an identity entry, but after an apply by key the new primary form had no entry at all.

## Fix

`apply_affine` no longer removes entries from `attrs["affines"]`. When you apply an affine by key, that key stays in the result as the identity (a `(pose, 4, 4)` stack of identity matrices for a per-pose affine). The identity is written exactly instead of computed as `arr @ inv(arr)`, which has floating-point error. The other entries are still re-expressed in the new world space.

## Change in `consolidate_poses`

Keeping the key broke a workflow that the `consolidate_poses` docstring recommends: call `da.fusi.affine.apply("<key>")` first, then `consolidate_poses`.

`consolidate_poses` requires every `(pose, 4, 4)` entry in `attrs["affines"]` to be a constant left-link of the primary geometry, `A[p] = L @ voxel_to_world[p]`. The applied key used to be gone by then, so the check never saw it. Now the key stays as a stack of identity matrices, which is not a left-link of a pose-dependent geometry, and `consolidate_poses` raised `ValueError`.

`_consolidate_affines` now handles this case before the left-link check: a `(pose, 4, 4)` entry that is the same for every pose collapses to that single `(4, 4)` affine. Consolidation does not move the world space, so a world-to-reference affine shared by all poses is still valid for the consolidated array. Entries that differ between poses still go through the left-link check.

## Tests

- `test_string_key_applied_kept_as_identity` replaces `test_string_key_applied_dropped_from_result`. It asserts that the applied key is the identity and that the other entry is re-expressed.
- `test_secondary_affine_rebase_before_consolidate` now also asserts that the applied key is the `(4, 4)` identity after consolidation.

## Documentation

- The `Returns` sections of `apply_affine` and `FUSIAffineAccessor.apply` say that the key stays as the identity.
- "Switching World Spaces" in `docs/user-guide/voxeldata.md` shows `world_to_qform` as the identity in the example output.
- The changelog has a new entry under `:bug: Fixes`.


docs/user-guide/voxeldata.md
```diff
@@ -436,10 +436,9 @@ array([[ 0., -1.,  0.,  0.],
 
 Applying `world_to_qform` absorbs the rotation into the DataArray's voxel-to-world
 affine, and the derived `z`/`y` coordinates change accordingly. The `qform` space
-becomes the new world space, and `"world_to_qform"` is dropped from the result:
-applying a stored affine by its own key re-anchors the world space to exactly that
-space, so the entry would carry no information any more. The `sform` entry, identity
-before, now holds the inverse rotation, so the `sform` space stays recoverable.
+becomes the new world space, so `"world_to_qform"` is now the identity. The `sform`
+entry, identity before, now holds the inverse rotation, so the `sform` space stays
+recoverable. Applying an affine never removes entries from `.attrs["affines"]`.
```
FelipeCybis:
```suggestion
recoverable.
```

FelipeCybis:
@sdiebolt we might have different views on this one. I do think it is important to have the ability to go back and forth from different reference spaces. One useful case is: let's say the qform is axis aligned and the sform is oblique, then I like to plot things in the qform space as it makes things a bit easier. Of course this is useful for the data exploration phase only.

docs/changelog.md
```diff
@@ -21,6 +21,12 @@ Released 2026-09-16.
   large recordings or broad noise masks
   ([#434](https://github.com/confusius-tools/confusius/pull/434)).
 
+### :bug: Fixes
+
+- [`apply_affine`][confusius.xarray.apply_affine] keeps the applied key in
+  `.attrs["affines"]` as the identity when the affine is given by key
+  ([#463](https://github.com/confusius-tools/confusius/pull/463)).
+
```
FelipeCybis:
```suggestion
## 0.7.2.dev0

Current development version for the next ConfUSIus release.

### :bug: Fixes

- [`apply_affine`][confusius.xarray.apply_affine] keeps the applied key in
  `.attrs["affines"]` as the identity when the affine is given by key
  ([#463](https://github.com/confusius-tools/confusius/pull/463)).

## 0.7.1

Released 2026-09-16.

### :zap: Performance

- [`compute_compcor_confounds`][confusius.signal.compute_compcor_confounds] no
longer computes a full SVD, extracting components several times faster on
large recordings or broad noise masks
([#434](https://github.com/confusius-tools/confusius/pull/434)).

```

sdiebolt:
> @sdiebolt we might have different views on this one. I do think it is important to have the ability to go back and forth from different reference spaces. One useful case is: let's say the qform is axis aligned and the sform is oblique, then I like to plot things in the qform space as it makes things a bit easier. Of course this is useful for the data exploration phase only.

Had some time to think about it during the holiday, and totally agree with you on this one!

sdiebolt:
Woops sorry, undrafted by mistake!

FelipeCybis:
> Woops sorry, undrafted by mistake!

No problem! In matter of fact, this PR is indeed ready for review now :)

